PR #17, opened 2018-09-24
Replace QML javascript with a python model

### LINKED ISSUE #16: Avoid javascript business logic
As discussed following #14 we want as close to all of the business logic of the application to be done in python instead of QML javascript so we're not introducing additional languages to the teams projects.

ACLay:
Closes #16


geometry-constructor/Models.py
```diff
@@ -0,0 +1,46 @@
+import attr
+from PySide2.QtCore import Qt, QAbstractListModel, QModelIndex, Slot
+
+
+@attr.s
+class Pixel:
+    name = attr.ib(str)
+    faces = attr.ib(factory=list)
```
matthew-d-jones: :+1: extra points for nice use of `attrs`

PR #21, opened 2018-09-26
Add a simple Scene3D animation of a neutron beam

### LINKED ISSUE #19: Try embeding a Scene3D in the interface
Hopefully this is as simple as adding a few lines to the QML...

ACLay:
Closes #19

image

### DIFF
--- a/geometry-constructor/Qt models/main.qml
+++ b/geometry-constructor/Qt models/main.qml
@@ -15,7 +15,7 @@ Rectangle {
     }
 
     Rectangle {
-        id: animationArea
+        id: instrumentViewArea
         anchors.top: parent.top
         anchors.bottom: parent.bottom
         anchors.left: pixelFieldsArea.right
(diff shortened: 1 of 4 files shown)


geometry-constructor/Qt models/main.qml
```diff
@@ -1,130 +1,163 @@
...
             }
-            filenameField.text = path
         }
     }
 
-    Logger {
-        id: myLogger
-    }
+    Rectangle {
+        id: animationArea
```
matthew-d-jones: Let's call this `instrumentViewArea`.

geometry-constructor/Qt models/main.qml
```diff
@@ -1,130 +1,163 @@
...
 import MyModels 1.0
 
 Rectangle {
-    width: 400; height: 300
+    width: 700
+    height: 300
 
-    Row {
-        id: textRow
+    Rectangle {
```
matthew-d-jones: `main.qml` is getting large, could you move everything which is in the main panel into its own qml file?

PR #35, opened 2018-10-03
Enable distributable versions of the app to be built

### LINKED ISSUE #27: Windows executable
Having a single file executable or installer early in development would make life easier for potential testers of the application.
`pyinstaller` and `pyupdater` are probably the standard tools for this at the moment, worth checking though.

ACLay:
Closes #27 

The app, tests, and distributable version all seem to be good on my windows machine.

The build tool it uses is cross platform, so @matthew-d-jones would you be able to check if the build distributable command (`python setup.py build_exe`) works on your linux box?


matthew-d-jones:
:+1: Working well on Linux too. See if excluding unused parts of Qt from the package is possible.

matthew-d-jones:
I think it is worth also getting rid of these:
```
"lib/PySide2/Qt/resources",
"lib/PySide2/Qt/translations"
```
You can `import shutil` to remove (non-empty) directories with `shutil.rmtree(full_path)`.

ACLay:
In total, this should now have removed about

- 180MB from 486MB on windows (37%)
- 249MB from 660MB on linux (38%)

from the frozen output directories

PR #36, opened 2018-10-05
26 config files

### LINKED ISSUE #26: Add a qtquickcontrols2.conf file to configure appearance
https://doc.qt.io/qt-5/qtquickcontrols2-configuration.html

ACLay:
Closes #26 

I realise that adding generated code (`geometry_constructor/resources.py` is built from the qrc resources file) to the repo may not be the cleanest solution, but doing so here would mean that to use/run the software from github doesn't require an additional build step.


matthew-d-jones:
Have you seen the `.conf` files: https://doc.qt.io/qt-5/qtquickcontrols2-configuration.html
Does the qrc method provide some functionality which the `.conf` does not?

If the qrc approach is necessary then this PR is good, thank you for the clear documentation.

ACLay:
To load the `.conf` files [they need to be included in a `.qrc`](https://doc.qt.io/qt-5/qtquickcontrols2-configuration.html#using-the-configuration-file-in-a-project) which gets compiled into the project. That page says to add it to the qmake build config `.pro` file to compile it, but for pyside2 I found the alternate method which I documented here.

PR #37, opened 2018-10-09
Real implementation window

### LINKED ISSUE #29: Create Window with main controls for real implementation
Start "real" implementation and rename existing directory to indicate it is the prototype.
Create an application window with the main controls: `Scene3D`, component `ListView`, add component button and `MenuBar`.

ACLay:
Closes #29 

Adapts the demo UI to have the basic elements required for the main screen of the actual app.
Greatly expands the model classes and allows it to be saved to a nexus file, though it doesn't store transformations at present.

image


resources/Qt models/main.qml
```diff
@@ -1,48 +1,113 @@
...
-        leftPadding: 0
 
-        width: 365;
+    menuBar: MenuBar {
+        Menu {
+            title: "File"
+            Action {
+                text: "Open"
+                enabled: false
+                onTriggered: myLogger.log("Hello World!")
```
matthew-d-jones: May as well put something useful in these log messages

tests/test_hdf_writer.py
```diff
@@ -0,0 +1,82 @@
+import h5py
+from geometry_constructor.Writers import HdfWriter
+from geometry_constructor.Models import InstrumentModel, Detector, PixelGrid, Corner, CountDirection
+
+
+def assess_pixel_grid_direction(direction, corner, expected_data):
```
matthew-d-jones: See if you can use the "in-memory" file option to avoid having to do disk IO in tests.
http://assorted-experience.blogspot.com/2015/06/h5py-in-memory-files.html

ACLay: Looks like to completely avoid disk writes we'd have to avoid closing the file in the save function, or it'd be dropped from memory. I don't think we should be leaving the file open, so don't think we should use the purely in memory option.

It looks like the hybrid option will still write to file, but keep a copy in memory which may reduce the number of disk ops. However then we'd be testing the code with a mode that it wouldn't be used in for the real app, unless you want the memory mapped mode to be used in production too?

resources/qtquickcontrols2.conf
```diff
@@ -1,2 +1,5 @@
 [Controls]
 Style=Material
+
+[Material]
+theme=Light
```
matthew-d-jones: Should be capitalised: `Theme`

prototype/resources/Qt models/AnimatedEntity.qml
```diff
@@ -0,0 +1,174 @@
...
+        viewCenter: Qt.vector3d( 0.0, 0.0, 0.0 )
+    }
+
+    FirstPersonCameraController { camera: camera }
+
+    components: [
+        RenderSettings {
+            activeFrameGraph: ForwardRenderer {
+                camera: camera
+                clearColor: "white"
```
matthew-d-jones: I think "lightgrey" looks good for this.

PR #48, opened 2018-11-13
Input validation

### LINKED ISSUE #43: Input validation and error display
Ensure that number fields are numeric, and that translate parent's don't form circular dependencies.

ACLay:
Closes #43 

- Adds validators to all the numeric input fields
- Adds custom validators to ensure unique component names and to prevent dependency loops in transform parents
  - Displays tool tips when custom validators fail
- Prevents removing components that are transform parents to other components

image

As the behaviour of text boxes with the built in validators seems to be to prevent invalid input, I've added similar code for the parent drop down. If an invalid option is clicked, it will go back to the previous selection.

### DIFF
--- a/geometry_constructor/validators.py
+++ b/geometry_constructor/validators.py
@@ -62,9 +62,13 @@ class TransformParentValidator(ValidatorOnInstrumentModel):
 
         The signal 'validationFailed' is emitted only if a circular dependency is found
         (http://doc.qt.io/qt-5/qvalidator.html#validate)
+        :param input: The name of the component being considered for assignment as a transform parent
+        :param pos: The cursor position in input. Required to match the QValidator API, and not used.
         """
-        parents = {}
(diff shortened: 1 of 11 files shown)


geometry_constructor/validators.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Validates the input as the name of a component's transform parent to check it won't cause a circular dependency
+
+        The signal 'validationFailed' is emitted only if a circular dependency is found
+        (http://doc.qt.io/qt-5/qvalidator.html#validate)
+        """
+        parents = {}
+
+        candidate_parent_index = -1
+        for component in self.instrument_model.components:
```
matthew-d-jones: Could you make this function easier to follow? What is the role of each of these `for` loops, could they be extracted to separate functions with elucidating names?

PR #50, opened 2018-11-14
Load geometry from STL files

### LINKED ISSUE #49: Consider supporting reading shape from STL or STEP format
The Design Division at ISIS uses "Solid Edge" CAD software which does not output to OFF or OBJ without an expensive plugin. It does output to STL or STEP, so if you estimate a day or less to implement support for one of these formats then it will be worthwhile, otherwise we'll manage with converting via an intermediate application for now.

The relevant (shape, not texture etc) information in such files is pretty much the same as in OFF: vertex coords and face definitions using vertex indices. My hope is therefore that it would be low effort to introduce support.

Looks like there is at least one suitable python library for reading STL files: https://pypi.org/project/numpy-stl/

ACLay:
Closes #49 

Also significantly improves update rendering time in the JSON display when using large geometry files, doing parent assignment and in an existing O(n) loop rather than a separate O(n<sup>2</sup>) loop


tests/test_geometry_loader.py
```diff
@@ -82,6 +82,12 @@ def test_load_stl_file():
         assert face_found
 
 
+def test_load_non_geometry_file():
```
matthew-d-jones: Try to use the test name to document as explicitly as possible what you are checking. In this case perhaps `test_load_geometry_returns_empty_geometry_for_unrecognised_file_extension`.

Issue #56, opened 2018-11-19
Single component visualisation in Edit Component Window


ACLay:
Being able to view the current component would be useful for knowing details have been entered correctly, and I feel essential for a useful face number to pixel id mapping interface.
However, due to a bug/bugs in Qt3D, it's not currently possible to have Scene3D instances in child windows.

Examples that seem to match my experiences trying to add a Scene3D to the editor window can be found in the Qt bug tracker at: 
- https://bugreports.qt.io/browse/QTBUG-69877
- https://bugreports.qt.io/browse/QTBUG-64974
- https://bugreports.qt.io/browse/QTBUG-70747

Some of these have linked code, one currently due to release with Qt 5.11.3, and Qt 5.12.0, so we should wait for the next release before tackling this issue

rerpha:
I suspect this is no longer an issue, technically the web view now uses an openGL context, and that is in a child window/dialog. I think this should be doable, perhaps we should put the 3d view as a pane in the geometry file dialog.

rerpha:
if we are going to do this we should keep in mind the solution in #432 as it may happen again to this and crash the program when closing the add component window.

PR #57, opened 2018-11-19
Full component editor

### LINKED ISSUE #47: Add full editor screen for already added components

ACLay:
Closes #47 

Adds controls for editing a detectors pixel information, and an editor window for already added components. Face to pixel mapping controls are impractically basic, but due to #56 and QtQuickControls 2 not having a TableView until Qt 5.12, I don't think much more can be done for that at present.

All the editor fields, both in the main and full editor windows now update the model automatically (when values are valid).

image

### DIFF
--- a/geometry_constructor/application.py
+++ b/geometry_constructor/application.py
@@ -6,7 +6,7 @@ Loading this module also registers with QML the required custom classes to load 
 
 import sys
 from os import path
-from geometry_constructor.instrument_model import InstrumentModel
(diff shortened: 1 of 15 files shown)


ACLay:
@matthew-d-jones I think that's all your feedback on this one addressed now.

ACLay:
Actually, this isn't ready yet. Found some bugs in my window changes that need sorting.

ACLay:
Right, I think this should be good now

PR #61, opened 2018-11-27
Expand range of available components

### LINKED ISSUE #32: Add component type selector dialog
Add component type selector dialog to "add component" button. I suggest "source" (see `NXsource` documentation) and "detector" options would be good to start with.

ACLay:
Closes #32 

Turns the "Add detector" window into an "Add component" window, adding Monitor, Source, Slit, Moderator, and Disk Chopper as options.

image

In the backend, a components type is now a property of Component, as the varying subclasses had so few distinguishing properties they didn't seem worth keeping. This also has the side effect of making Pycharm's type inferring much happier when dealing with component classes.


ACLay:
I feel like the pixel model classes have so little logic of their own, as they just implement the QML ListModel interface for their respective classes, that any tests would be a long repetitious way of saying "set property x to foo. does x equal foo?". It could be done, but I doubt it'd be very useful.

PR #102, opened 2019-02-11
Notify the user when their JSON is invalid or of the wrong type

### LINKED ISSUE #99: Notify the user when their JSON is invalid or of the wrong type
Currently the program throws an exception (is there handling for invalid json?) 

if there isn't handling for it, add some. 

Also notify the user that their file couldn't be parsed.

DolicaAkelloEgwel:
### Issue

Closes #99 

### Description of work

Informs user that the JSON file they provided was invalid.

The `load_file_into_instrument_model` in `JsonConnector` now returns a boolean indicating whether or not file reading was successful. This is received in `Main.qml` which uses this to decide if it's going to show a bad file message.

The tests in `test_json_connector.py` verify that `json_string_to_instrument_model` returns true for valid JSON and false for invalid JSON.

### Acceptance Criteria 

Attempt to load `empty.json` or `invalid.json` from `\\olympic\babylon5\Public\DolicaAkelloEgwel\NexusConstructor\`. In both cases, the program should show a message indicating that the file couldn't be parsed. For `valid.json` (also in the same directory) the file should load without showing the message.


nexus_constructor/json_connector.py
```diff
@@ -32,23 +32,32 @@ def __init__(self):
...
 
-    @Slot(QUrl, 'QVariant')
+    @Slot(QUrl, 'QVariant', result=bool)
     def load_file_into_instrument_model(self, file_url: QUrl, model: InstrumentModel):
         filename = file_url.toString(options=QUrl.PreferLocalFile)
         with open(filename, 'r') as file:
             json_string = file.read()
-        data = json.loads(json_string)
+
+        try:
```
matthew-d-jones: If this line onwards was extracted to a separate method then it could be unit tested by passing json strings.

tests/test_json_connector.py
```diff
@@ -0,0 +1,27 @@
+from nexus_constructor.json_connector import JsonConnector
+from nexus_constructor.qml_models.instrument_model import InstrumentModel as model
+
+
+def test_good_json_returns_true():
```
matthew-d-jones: I think valid/invalid is a little clearer than good/bad for the test names

tests/test_json_connector.py
```diff
@@ -0,0 +1,27 @@
...
+
+    json_connector = JsonConnector()
+
+    with open("tests/sample.json", mode="r") as file:
```
matthew-d-jones: It is good practice to avoid disk and network io in units tests, because we want to make them fast and independent of external systems (filesystem, network).

Instead of using files here you can simply create strings in the test. This has the advantage of the input being immediately visible to someone reading the test implementation. Creating it as a "raw string" using `r` in front of the quotes (`r"some string"`) will avoid needing to escape new line characters etc. Try to give shorter examples of the "good" and "bad" inputs.

PR #105, opened 2019-02-12
Update Qt

### LINKED ISSUE #75: Update Qt to 5.12
This should give us TableView in Qt Quick Controls 2, which may be useful. Hopefully also solve blocker for #56

DolicaAkelloEgwel:
### Issue

Closes #75 

### Description of work

Updates PySide and pytest.

### Acceptance Criteria 

Check that the tests still pass. Open some files to check that everything still works.


rerpha:
ok so the reason i think this isn't working is because centos requires `mesa-libGL.x86_64` to be installed for the newer versions of QT. I don't think this is an issue with ubuntu, so we could either move the build image to use ubuntu, or update the build node and make it install `mesa-libGL.x86_64` in the dockerfile. 

I think we should do both and get a parallel build working on both centos and ubuntu

rerpha:
Looks like we need python 3.6 for this. I have tried the red hat version and the tests pass.

PR #107, opened 2019-02-22
Unit selection dialog when loading geometry files

### LINKED ISSUE #51: Unit selection dialog when loading geometry files
OFF and STL files don't store any information about the units of their points (mm, cm, inches...).

DolicaAkelloEgwel:
### Issue

Closes #51 

### Description of work

Enables a user to select a unit of length when loading a geometry from a file or creating a cylinder. Adds additional text fields to the geometry-related menus for entering units. Uses `pint` library to manage checking if user input is a valid unit of length and converting between different units. Introduces a `UnitValidator` for ensuring valid units were entered. The validator is tested in `test_unit_validator` in `test_validators.py`. If the units given are not valid, a message is shown and the geometry cannot be created.

Meters are treated as the "default" units. This means that when they are selected as the units, the program multiplies the relevant values in a file/cylinder by 1. For cm they are multiplied by 0.01, for km 1000, etc.

I've also tried to add a bit about this to `DeveloperNotes.md`.

### Acceptance Criteria 

Load the sample geometry files/create a cylinder and try to:
- leave the units box blank - this shouldn't work
- enter gibberish in the units box - this shouldn't work
- enter a unit of volume/area/time/etc - this shouldn't work
- enter units with a magnitude such as "40 cm" - this should't work
- enter a number with no units - this shouldn't work
- enter m/meter/metre, yard/yards, foot/feet, etc - the program should be able to recognise when different inputs mean the same thing
- enter cm, press Add, translate the new geometry so that it isn't inside the red cube, and see check that the new geometry has 1/100th the height/length of the red cube (this may require zooming)
- inspect the JSON panel to the right and check that the points of any new geometries have the expected values
- enter invalid units, close the window/press cancel without creating a geometry, and then try to create a new geometry - this should cause the bad geometry message to disappear

Also have a look at the things I've added to `DeveloperNotes.md` to see if they make sense/are useful.

### DIFF
--- a/nexus_constructor/data_model.py
+++ b/nexus_constructor/data_model.py
@@ -69,10 +69,10 @@ class CylindricalGeometry(Geometry):
     described by the direction of its axis, its height, and radius.
     """
 
-    units = attr.ib(default="feet", type=str)
     axis_direction = attr.ib(factory=lambda: Vector(1, 0, 0), type=Vector, validator=validate_nonzero_vector)
     height = attr.ib(default=1, type=float)
     radius = attr.ib(default=1, type=float)
+    units = attr.ib(default="", type=str)
 
     @property
     def base_center_point(self):
--- a/nexus_constructor/data_model.py
+++ b/nexus_constructor/data_model.py
@@ -3,24 +3,7 @@ from enum import Enum, unique
 from math import sqrt, sin, cos, pi, acos
 from typing import List
 from PySide2.QtGui import QVector3D, QMatrix4x4
-import pint
-
-
-def calculate_unit_conversion_factor(units):
-    """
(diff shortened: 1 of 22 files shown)


nexus_constructor/data_model.py
```diff
@@ -50,6 +51,7 @@ class CylindricalGeometry(Geometry):
     The cylinder is assumed to have the center of its base located at the origin of the local coordinate system, and is
     described by the direction of its axis, its height, and radius.
     """
+    units = attr.ib(default="m", type=str)
```
DolicaAkelloEgwel: I've noticed one of the tests fails if the default unit is changed to anything other than metres. I'm not sure if this a problem or not.

PR #109, opened 2019-02-26
Make various text labels selectable

### LINKED ISSUE #106: Make various text labels selectable
The nexus structure should probably be able to be selected or there should at least be a button to copy it to the clipboard so it can be used in File writer commands for manual testing etc 

There are also some other fields, like component names and descriptions, which I can't think of a good reason as to why they shouldn't be selected. If everyone agrees we should make them select-able.

DolicaAkelloEgwel:
### Issue

Closes #106 

### Description of work

Allows the contents of the JSON panel to be copied to the clipboard by using the `pyperclip` library. 

Some differences when using "Copy to Clipboard":
- the keys/values are copied in a different order (in the window it looks like they are arranged in alphabetical order)
- lists values don't all appear on the same line like they do in the window

The content of the `TextField`s in the component window should now be selectable. Copying/pasting can only be done with the keyboard for the time being. 

Note: The functions for copying to the keyboard aren't tested as they are both only one line long and would require using copy/paste on Jenkins (this is responsible for the codecov % drop). The JSON to place in the clipboard is obtained from `generate_json()` which is already tested elsewhere.

### Acceptance Criteria 

Switch between the different JSON modes and check that pressing copy to clipboard does the right thing. Check that selecting "Hide JSON Display" hides the "Copy to Clipboard" button as well. Attempt to copy/paste things into the different `TextField`s but make sure this doesn't work for the Geometry file path input.


matthew-d-jones:
@DolicaAkelloEgwel PySide2 actually has clipboard functionality already in it:
https://doc.qt.io/qtforpython/PySide2/QtGui/QClipboard.html
It is worth trying if you haven't already, if it works as well as pyperclip then better not to have the extra dependency.

DolicaAkelloEgwel:
Is the information about `xclip` still needed when using `QClipboard`?

Issue #112, opened 2019-03-04
Transform in "add component" window doesn't seem to work 


rerpha:
Seems to work OK in the components pane, but for some reason when adding a component and setting a translation via the sample it logs an exception and refuses to add the component. 

```
Traceback (most recent call last):
  File "nexus-constructor/nexus_constructor/qml_models/instrument_model.py", line 362, in get_transform_model
    return self.transform_models[index]
IndexError: list index out of range
```

DominicOram:
May be fixed by https://github.com/ess-dmsc/nexus-constructor/issues/188

PR #113, opened 2019-03-05
Adding an option for no geometry and adding a dummy object

### LINKED ISSUE #62: Dummy object in 3D view for component with no shape
A "None" option should be available for shape of any component type. Selecting "None" would indicate that a dummy object should be placed in the 3D view at the position of the component, but that there would be no geometry for the component in the output file.

rerpha:
### Issue

Closes #62 

### Description of work

I have added a dummy object as well as a separate model for when the component has no shape. 

### Acceptance Criteria 
- Add a component, use the "None" button under geometry type, Check that a cube is displayed in the 3d view. 
- Translations should still work, same with rotations. Note that #112 exists so you'll need to transform after adding a component. 
- The outputted file should not contain any geometry for this object 


nexus_constructor/off_renderer.py
```diff
@@ -171,11 +170,42 @@ def __init__(
...
         self.setFirstVertex(0)
-        self.setFirstInstance(0)
         self.setPrimitiveType(Qt3DRender.QGeometryRenderer.Triangles)
+        self.setFirstInstance(0)
         self.setGeometry(qt_geometry)
-        self.setVertexCount(qt_geometry.vertex_count)
+
+    def create_dummy_object(self):
+        """
+        Create a dummy OFF geometry that displays as a square for when the component has no geometry.
```
matthew-d-jones: "cube" not "square"

nexus_constructor/data_model.py
```diff
@@ -195,6 +195,13 @@ def winding_order_indices(self):
         return [sum(face_sizes[0:i]) for i in range(len(face_sizes))]
 
 
+@attr.s
+class NoShapeGeometry(Geometry):
+    """
+    Dummy object for components with no geometry - sphere
```
matthew-d-jones: remove "- sphere"

PR #125, opened 2019-03-07
Use in-memory nexus file for model

### LINKED ISSUE #78: Remove non-standard json description
We can currently output to our NeXus-json description compatible with the file writer, and another json description. The latter of these should be removed. It may be in use because the NeXus-json description is lossy, in which case may need to wait until we have a non-lossy save format (NeXus file?) before removing it.

See "JSON Formats" section of README, and make sure README is updated as part of this ticket.

### LINKED ISSUE #91: Use NeXus as the model
I propose changing the models to use NeXus via  the nexusformat library: www.github.com/nexpy/nexusformat
Basically this would be an in-memory representation of the NeXus file. This change would make the save and load functionality trivial.

Some data may need to exist in a separate model/models, for example the meshes for 3D visualisation?

This is the best documentation for how to use nexusformat:
https://github.com/nexpy/nexusformat/blob/master/src/nexusformat/notebooks/nexusformat.ipynb
`nexusformat` looks like it should be more convenient to use than h5py, but if there is an unforeseen problem with using it then using h5py directly would also be a good option.

### LINKED ISSUE #108: Use in-memory NeXus for part of model
Test idea of #91 
Identify a model, or part of model to reimplement using NeXus.
Use h5py, with in-memory option for file. See the `'core'` driver option: http://docs.h5py.org/en/stable/high/file.html

The transformation model is a good candidate. A first step would be creating an in-memory NeXus file in the transformation model and reading/writing the transformation properties to that.

The idea eventually would be to have a single NeXus file which all models store their data in, as this will make file loading and writing easy. To prove this approach would be viable the next step would be to create the in-memory NeXus file outside of the transformation model, pass it in the constructor and the model constructor would add a new NXtransformation to the file.

### LINKED ISSUE #124: Have Transformation be an abstract class
It doesn't seem to be instantiated anywhere. Making it an abstract class might be good for preventing accidents.

DolicaAkelloEgwel:
### Issue

Closes #108 
Closes #124 
Closes #78 
Closes #91 

### Description of work

Note: this is a long-running PR for converting the model to use an in-memory nexus file. We should keep merging master into this so it doesn't get outdated. 

Sets up the `Transformation` class and its subclasses to use in-memory Nexus file. This causes issues with multiple `Transformation`s having the same name. As a result, some tests were changed so the transformations have unique names. Tests involving converting to/from a JSON file are now skipped (this might stop being a problem once a single NeXus file is used for the entire model). The `Transformation` classes now have their own file and some matching tests.

### Acceptance Criteria 

Add translations and rotations to an object and check that they work correctly.

### DIFF
--- a/nexus_constructor/transformation.py
+++ b/nexus_constructor/transformation.py
@@ -24,7 +24,7 @@ class Transformation:
 
 
 class Rotation(Transformation):
-    def __init__(self, angle: int, name="rotation", axis=Vector(0, 0, 1)):
+    def __init__(self, angle=0.0, name="rotation", axis=Vector(0, 0, 1)):
         """
         Creates a rotation in the in-memory Nexus file under
         :param angle: The angle to rotate the object by.
@@ -34,12 +34,16 @@ class Rotation(Transformation):
         super().__init__(name)
         self.transformation.attrs["transformation_type"] = "rotation"
         self.transformation.attrs["angle"] = angle
-        self.transformation.attrs["axis"] = axis.xyz_list
+        self.transformation.attrs["axis"] = axis.vector.tolist()
         self._axis = axis
 
     @property
     def angle(self):
         return self.transformation.attrs["angle"]
+
+    @angle.setter
+    def angle(self, angle):
+        self.transformation.attrs["angle"] = angle
 
     @property
     def axis(self):
@@ -55,7 +59,7 @@ class Translation(Transformation):
         """
         super().__init__(name)
         self.transformation.attrs["transformation_type"] = "translation"
-        self.transformation.attrs["vector"] = vector.xyz_list
+        self.transformation.attrs["vector"] = vector.vector.tolist()
         self._vector = vector
 
     @property
--- a/nexus_constructor/transformation.py
+++ b/nexus_constructor/transformation.py
@@ -1,0 +1,67 @@
+import h5py
+from nexus_constructor.vector import Vector
+
+
+class Transformation:
+    """
+    Parent class for transformations on the model.
+    """
+
+    def __init__(self, name):
+        """
+        Creates an in-memory nexus file with a named entry containing an NXTransformation attribute.
+        :param name: The name of the root entry in the nexus file.
+        """
+        self.nexus_file = h5py.File(name, driver='core', backing_store=False)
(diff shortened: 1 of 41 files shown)


DolicaAkelloEgwel:
Just now realising it's a bit weird to not add angle and axis information to the NeXus file as well. Reminding myself to work on this tomorrow...

rerpha:
I have disabled some unit tests for this, but seeing as those tests include loading several files from the hard drive as well as containing logic I think we should replace them anyway. Tests have been added for `Transformation.py` so coverage in the added functionality has increased.

nexus_constructor/transformation.py
```diff
@@ -0,0 +1,63 @@
...
+        Creates an in-memory nexus file with a named entry containing an NXTransformation attribute.
+        :param name: The name of the root entry in the nexus file.
+        """
+        self.nexus_file = h5py.File(name, driver="core", backing_store=False)
+        self.transformation = self.nexus_file.create_group(name)
+        self.transformation.attrs["NX_class"] = "NXtransformation"
+
+    @property
+    def name(self):
+        return self.transformation.name[1:]
```
matthew-d-jones: Why do you drop the first character?

nexus_constructor/transformation.py
```diff
@@ -0,0 +1,63 @@
...
+    Parent class for transformations on the model.
+    """
+
+    @abstractmethod
+    def __init__(self, name):
+        """
+        Creates an in-memory nexus file with a named entry containing an NXTransformation attribute.
+        :param name: The name of the root entry in the nexus file.
+        """
+        self.nexus_file = h5py.File(name, driver="core", backing_store=False)
```
matthew-d-jones: Rather than disabling tests, you could use a uuid here instead of `name` as the name of the file.
```python
import uuid
filename = str(uuid.uuid4())
```

nexus_constructor/transformation.py
DolicaAkelloEgwel: It returns the full path which contains forward slashes.

rerpha:
Seeing as master and this branch have diverged, I am going to close this PR for now. We can create another when we're a bit further along and when everything is tested nicely.

PR #141, opened 2019-03-18
Add scrollbar to components pane

### LINKED ISSUE #131: Add scrollbar to components pane
Currently the components pane on the left side of the GUI is not obviously scrollable. Add a scrollbar to indicate that it is when lots of components are added. could be invisible if there are a small amount of components and no need to scroll.

DolicaAkelloEgwel:
### Issue

Closes #131 

### Description of work

Adds a scrollbar to the components pane that only appears when the content is larger than the pane. Also disappears when components/translations are removed and the contents are now small enough to fit inside the pane.

### Acceptance Criteria 

Add/remove some components/translations and check that the scrollbar becomes visible or invisible when expected.


rerpha:
Looks good, I think I have found a bug though, but don't think this has introduced it. 
new issue for it is #150

DominicOram:
This works OK for when I'm adding lots of components but the scrollbars don't appear/disappear when changing the height of the application. Also, is it possible to do this with the ScrollBar.AsNeeded policy rather than do the display logic yourself?

DolicaAkelloEgwel:
> This works OK for when I'm adding lots of components but the scrollbars don't appear/disappear when changing the height of the application.
Just noticed that yesterday afternoon. Trying to work on it...

>Also, is it possible to do this with the ScrollBar.AsNeeded policy rather than do the display logic yourself?
That makes the scrollbar only appear when the mouse hovers over the `Flickable` area and I figured that didn't resolve the issue of these things not being "obviously scrollable."

DominicOram:
Ok, how about:
```
                ScrollBar  {
                    id: bar
                    policy: ScrollBar.AsNeeded
                    active: true
                    onActiveChanged: {
                        if (!active)
                            active = true;
                    }
                }
```
You'll need to put a comment in as to why you've got the on active bit though but does that do what you wanted?

DolicaAkelloEgwel:
Yeah, that's helpful. Thanks.

DolicaAkelloEgwel:
I think this is all fixed up now if anyone wants to take a look...

DolicaAkelloEgwel:
>Given that we're using the same isActive pattern on #147 is there any way that we can create a standard scrollbar object that contains this code that both of them use?

I've just tried adding that. What do you think?

DominicOram:
I think it's better cos we can then modify #147 to use the same code. What do you think?

DolicaAkelloEgwel:
I think it's a swell idea.

mattclarke:
cos? <shakes head>

PR #164, opened 2019-03-21
135 notify user when geometry file is empty or invalid

### LINKED ISSUE #135: Notify user when geometry file is empty or invalid
When adding a component with a mesh, if the file is blank, clicking add does nothing. Highlight the field much like the unit validator part of the cylinder component view that warns the user with red text that they have no geometry file selected or that it is not a valid file.

DolicaAkelloEgwel:
### Issue

Closes #135

### Description of work

Introduces a `GeometryFileValidator` for checking that a file is a valid STL/OFF geometry. This is tested in the file geometry_file_validator.py. After loading an invalid file, a pop-up will appear asking the user to select another file. 

### Acceptance Criteria 

Attempt to load some of the files in `smb://olympic/babylon5/Scratch/dolica` and check that you get an error message. Also check that the unit selection dialog _doesn't_ appear and that the program doesn't attempt to make a geometry object with the invalid file.

Also try to press "Add" when a geometry file hasn't been selected. This should show a message. Select a file and the message shouldn't appear when "Add" is pressed again.

I am not so sure about the way the program behaves when you load a bad file then close the Add Geometry dialog and reopen it. Ideally that should "reset" everything and not cause any lingering bad file stuff to be present but I'm not 100% it does this. I would appreciate someone checking this to make sure.


resources/Qtmodels/GeometryFileSelected.qml
```diff
@@ -0,0 +1,5 @@
+pragma Singleton
+import QtQuick 2.0
+QtObject {
+    property bool geometryFileSelected: false
+}
```
matthew-d-jones: Missing new line at end of file

resources/Qtmodels/qmldir
```diff
@@ -1,2 +1,3 @@
 # qmldir
-singleton ValidUnits ValidUnits.qml
\ No newline at end of file
+singleton ValidUnits ValidUnits.qml
+singleton GeometryFileSelected GeometryFileSelected.qml
```
matthew-d-jones: Missing new line at end of file

resources/Qtmodels/GeometryControls.qml
```diff
@@ -66,17 +71,38 @@ Pane {
                 anchors.verticalCenter: fileTextField.verticalCenter
                 anchors.right: parent.right
                 text: "Choose file"
-                onClicked: filePicker.open()
+                onClicked: {
+                    console.log("Opening file picker.")
```
rerpha: these should probably be removed if they are debug statements?

DolicaAkelloEgwel: Oops!

PR #166, opened 2019-03-21
Removing vector and replacing with QVector3D in transformations

### LINKED ISSUE #126: Replace Vector with QVector3D in Transformation classes
Wait for #120 and #108 to be complete before dealing with this.

rerpha:
### Issue

Closes #126 

### Description of work

Completely removing the `Vector` class and replacing with QVector3D.

In theory this should speed translations up, but we don't really have a good way of measuring this. 

I think we should start converting from roles and qabstractlists in a separate PR, hence why i have left the roles stuff in. 

### Acceptance Criteria 
Not really any functional changes here besides swapping calls to magnitude and normal, also there's a bit of code that has changed in the transform model but i think this should be a bit more readable. 


tests/test_instrument_model.py
```diff
@@ -183,12 +183,12 @@ def test_generate_matrix_combines_dependent_transforms():
...
     def rotate_matrix(matrix: QMatrix4x4, rotate: data_model.Rotation):
-        matrix.rotate(
-            rotate.angle, QVector3D(rotate.axis.x, rotate.axis.y, rotate.axis.z)
-        )
+        matrix.rotate(rotate.angle, rotate.axis)
 
     def translate_matrix(matrix: QMatrix4x4, translate: data_model.Translation):
-        matrix.translate(translate.vector.x, translate.vector.y, translate.vector.z)
+        matrix.translate(
+            translate.vector.x(), translate.vector.y(), translate.vector.z()
```
DolicaAkelloEgwel: Can you pass the vector here?

PR #170, opened 2019-03-22
Moving helper functions into helpers.py, removing calls in __init__ files.

### LINKED ISSUE #146: __init__.py files contain functions
Several of the `__init__.py` files in the project contain functions which are not used explicitly when the package is used. I think this is bad practice (involves importing code that you might not need to use) and the code should be moved out into somewhere sensible.

rerpha:
### Issue

Closes #146 

### Description of work

No functional changes, just moving some code out of the `__init__.py` files that don't need to be there. 

### Acceptance Criteria 

Code should be more pythonic, and easier to test the helpers. 


rerpha:
in progress, adding some tests.

tests/test_qml_model_helpers.py
```diff
@@ -0,0 +1,56 @@
+from nexus_constructor.qml_models.helpers import change_value, generate_unique_name
+from nexus_constructor.data_model import Component, ComponentType
+
+
+def test_GIVEN_different_attribute_WHEN_change_pv_value_called_THEN_changes_attribute_to_new_value():
```
matthew-d-jones: I think the test naming is a little strange, "pv" doesn't seem relevant?

PR #183, opened 2019-04-04
Enable MacOS executable build

### LINKED ISSUE #88: support Mac OS and create build in jenkinsfile

rerpha:
### Issue

Closes #88 

### Description of work

Enabling the MacOS build in the jenkinsfile to archive the executables for MacOS

### Acceptance Criteria 

Test that the executable runs on MacOS without any system libraries installed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -166,7 +166,7 @@ node("docker") {
         def builders = [:]
         builders['centos7'] = get_linux_pipeline()
         // disabled for now as the build isn't setup for Mac OS just yet.
-        // builders['macOS'] = get_macos_pipeline()
+        builders['macOS'] = get_macos_pipeline()
         builders['windows10'] = get_win10_pipeline()
 
         parallel builders


rerpha:
No idea what's going on here, @SkyToGround if you can get the executables working on macos please let me know otherwise it might be a bug in pyside2.

SkyToGround:
I have previously (repeatedly) tried to package Python application into binaries for all three major platforms (Linux, Windows, MacOS). I have never gotten it to work properly on MacOSX. I can take a look at it but for now I am fine with ignoring the problem. Especially if it as hard to solve as I think it is.

rerpha:
> I have previously (repeatedly) tried to package Python application into binaries for all three major platforms (Linux, Windows, MacOS). I have never gotten it to work properly on MacOSX. I can take a look at it but for now I am fine with ignoring the problem. Especially if it as hard to solve as I think it is.

fair enough. If you have no luck we'll just run the tests on macos i think

Issue #152, opened 2019-03-20
Investigate a better method for pane width/height


DominicOram:
Throughout the QML the is a pattern like the following on a number of panes:
```
contentHeight: typeLabel.height +
                           typePane.height +
                           geometryLabel.height +
                           geometryPane.height +
                           pixelLabel.height +
                           pixelPane.height +
                           continueButton.height
contentWidth: Math.max(typeLabel.width,
                           typePane.width,
                           geometryLabel.width,
                           geometryPane.width,
                           pixelLabel.width,
                           pixelPane.width,
                           250)
```
Manually calculating widths/heights like this looks messy and is error prone. QML most likely provides a better way to do this, which may be through [layouts ](https://doc.qt.io/qt-5/qtquicklayouts-index.html). Investigate other options and do the following:

* Modify the code to remove the pattern above wherever possible
* Document the best way to do this for future use
* The functionality of the UI should only change minimally due to this ticket.

PR #184, opened 2019-04-04
152 layouts for add component

DolicaAkelloEgwel:
### Issue

Makes some progress on #152 but doesn't close it.

### Description of work

- ComponentControsl.qml now has Layouts
    - `componentDelegate` has its own file, also uses Layouts
- AddComponentWindow.qml has Layours for the `setupPane`
    - `detailsPane` has its own file but this does not use Layouts
- Main.qml now uses a `RowLayout`
    - `jsonPane` has its own file. It was already using Layouts.
- TransformControls.qml has been broken down a little bit
    - `translatePane` has its own file
    - `rotatePane` has its own file
    - `transformButtons` has its own file

### Some issues I ran into

- When trying to tidy up Main.qml I noticed placing  `instrumentViewArea` in its own file leads to problems because it needs access to the `InstrumentModel` that is created in Main.qml

### Acceptance Criteria 

Use the Add Component Window and make sure it looks mostly the same as before. Also check the Components pane and "Add [Detector Type]" windows (the things that appear after pressing Continue) look the same.


rerpha:
Have tried playing around with instrumentViewArea and I can't get it to work outside of main.qml either. 
Let's leave it there for now

Issue #153, opened 2019-03-20
Move away from QAbstractListModel


DominicOram:
Most of the models inherit from QAbstractListModel and hold their properties as items within a list. This leads to:
* Some odd "magic numbers":
```
NameRole = Qt.UserRole + 1
DescriptionRole = Qt.UserRole + 2
TransformParentIndexRole = Qt.UserRole + 3
DependentTransformIndexRole = Qt.UserRole + 4
```
* Hard to understand logic for getting/setting the data
* Methods such as rowCount and roleNames, which are unclear to their purpose

This ticket is to:
* Find a more readable way to do this, maybe using [QObjects](https://doc.qt.io/qt-5/qobject.html)
* Document this pattern
* Add unit tests where appropriate
* Functionality of the program should generally not change

rerpha:
PRs are in for geometry models and (all but one) pixel models. The one that isn't a QObject is still an abstract list but I think that's the appropriate type for it. 

I can't figure out how to use properties correctly with abstract list, there's not any documentation on whether we should be using roles for that or properties. 

I think the transform model should also stay as a list as it contains a list of transforms. Likewise with instrumentModel, which contains a list of the components.

PR #190, opened 2019-04-09
Changing Geometry models to use QObject rather than QAbstractList

### LINKED ISSUE #169: Selecting another geometry file leads to a segfault
It looks like attempting to load a geometry file, entering its units, and then choosing another geometry file rather than pressing "Add" causes the program to crash.

### LINKED ISSUE #181: Qt errors from GeometryControls.qml
I get these frequently when using the nexus constructor: 
```
file:///home/jack/code/repos/nexus-constructor/resources/Qtmodels/GeometryControls.qml:272:21: Unable to assign PySide::PyObjectWrapper to QString
file:///home/jack/code/repos/nexus-constructor/resources/Qtmodels/GeometryControls.qml:281:21: Unable to assign PySide::PyObjectWrapper to QString
file:///home/jack/code/repos/nexus-constructor/resources/Qtmodels/GeometryControls.qml:290:21: Unable to assign PySide::PyObjectWrapper to QString
```
along with various other similar errors.

It looks like these are caused due to the way in where we have wrapped `Button`s and `TextField`s in QML, in the alias properties that pass the text through to the qml object (such as Button)

https://github.com/ess-dmsc/nexus-constructor/blob/c5bab5786b42fbf981a1308b9881493c5ddf6051/resources/Qtmodels/LabeledTextField.qml#L7

https://github.com/ess-dmsc/nexus-constructor/blob/c5bab5786b42fbf981a1308b9881493c5ddf6051/resources/Qtmodels/PaddedButton.qml#L6

I think we should either:
- Remove both of these classes, as it makes the code more simple and easy to read, or
- fix the alias type in both places

By removing, we do have duplication, but seeing as we are moving to grid and row layouts soon anyway I don't think layout should be handled here. I would rather just be able to use Qt's built in types from quickcontrols. It does mean that we might have to change the padding in several different places but it also means we have good QML documentation and no pointless wrapper types!

rerpha:
### Issue

Part of #153 

Closes #169 
Closes #181 

### Description of work

Not working just yet, need some help on it if possible please! everything but the OFF models work - when clicking "add" after adding a mesh component given a valid file it does nothing currently.

Changing the models in `geometry_models.py` to use a `QObject` rather than `QAbstractList` as there is no reason to use `QAbstractList`s. With a `QObject` we can add properties and remove the code that goes through `Qt.UserRole`s which makes the code a lot cleaner. 

Sadly getters and setters need to be used here (not pythonic), but I think its still a whole lot clearer what's going on than before

### Acceptance Criteria 
- tests all pass
- components can still be added
- transforms on components still works
- off and stl files still work 


rerpha:
todo: build shiboken2 and use qt creator to see if there is a qml syntax error. Likely there is, for some reason these are not being shown in the log anymore! at least not on windows

DominicOram:
As discussed, I believe this works OK but I think we can clean up some of the getter/setter boiler plate. Potentially using something similar to http://code.activestate.com/recipes/577539-convert-pyqt-properties-to-python-properties/. We should look into this briefly but timebox it to a day or so.

matthew-d-jones:
@DominicOram 
http://code.activestate.com/recipes/577539-convert-pyqt-properties-to-python-properties/ looks sensible, but are we going to need getters/setters anyway when the model is storing the properties in NeXus?

PR #192, opened 2019-04-11
175 replace tests that load from off stl and json files

### LINKED ISSUE #175: Replace tests that load from off, stl and json files
Bad idea because it means the unit tests rely on hardware i/o which is not guaranteed to always work. it also slows down the tests dramatically, some taking more than 2 seconds to run, which is a long time for unit tests. 

We should remove them when our code covers what they do, minus the loading/writing from json bit cos it'll be removed anyway.

DolicaAkelloEgwel:
### Issue

Closes #175

### Description of work

Removes file-loading from tests. I used MeshLab to convert [cube.stl](https://github.com/ess-dmsc/nexus-constructor/blob/master/tests/cube.stl) to an [ASCII file](https://github.com/ess-dmsc/nexus-constructor/blob/175_Replace_tests_that_load_from_off_stl_and_json_files/tests/asciicube.stl) so that it can be stored in a string and converted to `StringIO`. It's also in the repo just in case but maybe someone wants to remove it?

I've added the functions `load_file_to_geometry` (in OFFModel) and `_load_geometry` (in geometry_loader.py) that accept files rather than filenames as input. I've put these new functions in the tests in place of calls functions that take a filename. As a result of this change `load_geometry` now needs to catch a `UnicodeDecodeError` and attempt to open the file again in binary (this is usually something that the `stl` library does for you but this time it's being given a file rather than a filename). 

Some of the tests for comparing JSON now call `.strip()` because without this the tests fail if the indentation isn't identical.

### Acceptance Criteria 

Load different geometry files to make sure they can all be read like before. Check that all tests pass.


DominicOram:
Are you still doing work on this or do you want me to review it?

nexus_constructor/qml_models/geometry_models.py
```diff
@@ -176,6 +177,9 @@ def load_data(self):
         self.endResetModel()
         self.meshLoaded.emit()
 
+    def load_file_to_geometry(self, file, extension):
```
rerpha: I'm not sure we should add a function here just for testing.

PR #193, opened 2019-04-12
188 fix the logic of the change value method to take into account functions

### LINKED ISSUE #188: Fix the logic of the change_value method to take into account functions
The change_value function (https://github.com/ess-dmsc/nexus-constructor/blob/426824717d4b212fafbcd6044387e8dfc210f63b/nexus_constructor/qml_models/__init__.py#L4) is used in all the Qt Models when data is changed within them. It is responsible for changing the value on the underlying object and sending a signal about the change. However, it does not always work correctly as it assumes what is changing is a parameter and in some cases a method needs to be called. This is particularly obvious when we switched to using QVector3D, rather than changing x/y/z by calling ```vector.x = 10``` we must now call ```vector.setX(10)```. As such this code fails silently. This ticket is to:
* Make the code fail explicitly if the parameter does not already exist and so cannot be set like ```vector.x = 10```
* Fix the places where this happens, the best fix here may be not to use the method at all for these cases

Ultimately this method should probably be removed as part of https://github.com/ess-dmsc/nexus-constructor/issues/159 and https://github.com/ess-dmsc/nexus-constructor/issues/153 but this is not part of this ticket.

DolicaAkelloEgwel:
### Issue

Closes #188

### Description of work

Made some changes so that the vector values are accessed via a function call rather than everything just assuming it's an attribute. An error also get's thrown in `change_value` if `getattr` returns a function rather than an attribute.

### Acceptance Criteria 

Make sure tests pass and that transformations still do the right thing. Also changing axis direction for cylinders should work now.


rerpha:
#195 has been merged, this function is now in `instrument_model.py`

DolicaAkelloEgwel:
@rerpha I think I've updated this now.

PR #195, opened 2019-04-16
Making pixel models inherit from QObject rather than QAbstractListModel

rerpha:
### Issue

Relates to #153 

### Description of work

Refactored 2 of the pixel models that are wrapping pixel mapping types to inherit from `QObject` and use `Properties` 

- I tried using lambdas for the getters/setters, but there's no easy way as `self` is out of scope when declaring the `Property`.
- `PixelMappingModel` still uses a `QAbstractListModel`, however I think this is the appropriate thing to use in this scenario, at least for now until we figure a bit more about how to use normal `QObject`s with stuff like `ListView`s 
 - `change_value` is no longer used in this file. hooray! 

Have added some tests to cover the new changes, note that not all of the getters/setters are tested nor do they really need to be, as they don't cover much logic. 

### Acceptance Criteria 
To test these changes: 
`SinglePixelModel`:
- Add a component, choose `monitor` type and enter a pixel id that is not 0
- Check in the JSON output that this has been reflected.

`PixelGridModel`
- Add a component, choose `detector` and `Repeatable grid`. 
- Use the `cube.off` file in `tests/` for the geometry file
- change the pixel grid to use something like 2 columns and 3 rows, check that the shape is a rectangle that is 2 wide and 3 high.
- repeat but test with row heights and widths 
- repeat and check that pixel ids are reflected in the JSON 

`PixelMappingModel` has not changed in any way so no need to test this unless you really want to. 


tests/test_pixel_models.py
```diff
@@ -0,0 +1,102 @@
...
+    pixel_data = PixelGrid(rows=5)
+    model = PixelGridModel()
+    inst = InstrumentModel()
+    inst.components[0].pixel_data = pixel_data
+    model.set_pixel_model(0, inst)
+
+    assert pixel_data == model.get_pixel_model()
+
+
+def test_GIVEN_another_pixel_id_object_WHEN_setting_model_THEN_model_is_replaced():
```
matthew-d-jones: test name doesn't seem to match implementation

tests/test_pixel_models.py
```diff
@@ -0,0 +1,102 @@
...
+    model = SinglePixelModel()
+    model.set_pixel_id(custom_id)
+    assert model.get_pixel_id() == custom_id
+
+
+def test_GIVEN_pixel_model_WHEN_creating_pixelgridmodel_THEN_model_is_updated():
+    pixel_data = PixelGrid(rows=5)
+    model = PixelGridModel()
+    inst = InstrumentModel()
+    inst.components[0].pixel_data = pixel_data
```
matthew-d-jones: Don't assume inst.components[0] exists or that you are allowed to set pixel_data on it. Add a new detector component and set the pixel_data on that.

nexus_constructor/qml_models/pixel_models.py
```diff
@@ -160,7 +140,7 @@ def get_pixel_model(self):
     @Slot(int, "QVariant")
     def set_pixel_model(self, index, instrument: InstrumentModel):
         component = instrument.components[index]
-        if isinstance(component.pixel_data, PixelMapping):
+        if component.pixel_data.type == "PixelMapping":
```
matthew-d-jones: @DominicOram Jack says switching from using `isinstance` to having a string type field is a change you suggested. Is that right? Why do you prefer that?

DominicOram: I don't remember suggesting that. Sorry @rerpha, I probably hadn't had enough coffee, `isinstance` is better.

PR #235, opened 2019-05-02
214: Using StackLayout for Add Component

### LINKED ISSUE #214: Using a StackLayout for the Add Component Window
https://doc.qt.io/qt-5/qml-qtquick-layouts-stacklayout.html

Seems it would be much easier to understand what is happening by doing this rather than setting panes to visible/invisible. Should also add something to the Wiki.

DolicaAkelloEgwel:
### Issue

Closes #214

### Description of work

Uses a `StackLayout` to manage between switching between the `setupPane` and `detailsPane` in AddComponentWindow.qml.

### Acceptance Criteria 

Create different geometry types and press continue. Check that you arrive at the expected menu.

### UI tests

No intentional changes to UI behavior.


resources/Qtmodels/AddComponentWindow.qml
```diff
@@ -28,315 +28,309 @@ ExpandingWindow {
         anchors.fill: parent
         padding: 0
 
-        Pane {
-            id: setupPane
-            height: childrenRect.height
-            width: childrenRect.width
+        StackLayout {
+            id: stack
```
rerpha: this is just my opinion, might be wrong or you might disagree with it, but i think we should call this something a bit more descriptive rather than just stack. Perhaps something like `addComponentStack`? The reason i say this is that its declarative so anything can use `stack` and it might become a bit hard to find from elsewhere in the qml, and could cause confusion if we use another StackLayout

rerpha:
here is what i mean, this is a component with one translation added: 

Selection_049

as you can see the list of transforms is not visible until i resize the window: 

Selection_050

PR #238, opened 2019-05-03
231: Make sure checkFirstEnabled is called whenever the geometry type changes

### LINKED ISSUE #231: Disable Continue button in Add Component menu when a pixel layout hasn't been chosen
Or automatically select a valid pixel layout.

How to get the problem:
1. Click Add Component
2. Select "None" as the Geometry type
3. Change the Component Type to Monitor
4. Change the Geometry type to Mesh/Cylinder

It's still possible to click Continue which then prevents certain parts of the menu from appearing that depend on radio buttons being pressed. And after that I was able to click Add even with the missing information.

DolicaAkelloEgwel:
### Issue

Closes #231

### Description of work

Makes sure that the `checkFirstEnabled` function is called whenever the geometry type changes. Previously it only did this when the component type changed so sometimes it was possible to press continue without setting a pixel type.

### Acceptance Criteria 

Make sure you don't get the problem I described in #231 or anything similar. Try it out with different component types and see that you don't get the same issue.

### UI tests

Didn't write anything as I view this as a fix rather than introducing something new. I can add something if people still want me to though?

### DIFF
--- a/resources/Qtmodels/AddComponentWindow.qml
+++ b/resources/Qtmodels/AddComponentWindow.qml
@@ -79,6 +79,7 @@ ExpandingWindow {
                             onClicked: {
                                 setupPane.geometryState = "OFF"
                                 GeometryFileSelected.geometryFileSelected = false
+                                pixelPane.checkFirstEnabled()
                             }
 
                             checked: true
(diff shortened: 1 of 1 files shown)


rerpha:
was it actually possible to not select a pixel type? On master there always seems to be one checked anyway even when all of them are disabled

PR #239, opened 2019-05-03
232: Component name error message is duplicated in QML

### LINKED ISSUE #232: "Component names must be unique" message is duplicated in QML
Maybe it can go in an error message singleton along with the bad units message.

DolicaAkelloEgwel:
### Issue

Closes #232

### Description of work

Created a singleton for storing error messages. Now the bad units message has been move there too. The places where the message appeared have been replaced with `ErrorMessages.repeatedComponentName`.

### Acceptance Criteria 

Attempt to create multiple components with the same name and check that the message appears as before. Do the same with incorrect units to check that this also still works.

### UI tests

No intentional changes made to UI behavior.


resources/Qtmodels/ErrorMessages.qml
```diff
@@ -0,0 +1,6 @@
+pragma Singleton
```
rerpha: Does this need to be a singleton? do we care about instantiation count? If not and if QML allows this I think we should remove this declarator? and possibly do the same for the other variables that are declared as singletons

PR #267, opened 2019-05-08
261: StackLayout for TransformControls

### LINKED ISSUE #261: StackLayout for TransformControls
Assuming it can be done I think it would be nice to have a `StackLayout` for managing whether the Translation/Rotation items appear. This is how it's currently done:

```qml
states: [
    State {
        name: "Translate"; when: transform_type == "Translate"
        PropertyChanges { target: rotatePane; visible: false }
        PropertyChanges { target: rotatePane; height: 0 }
    },
    State {
        name: "Rotate"; when: transform_type == "Rotate"
        PropertyChanges { target: translatePane; visible: false }
        PropertyChanges { target: translatePane; height: 0 }
    }
]
```

DolicaAkelloEgwel:
### Issue

Closes #261 

### Description of work

Uses a `StackLayout` to manage the switch between the rotation stuff and the transformation stuff. The `StackLayout` is also bundled into a `ColumnLayout` with `transformButtons` (where the move up/down and delete buttons are kept). The result is that the height of the surrounding pane can now have its size set by using the height/width of `transformBoxColumn` instead of adding things up.
```qml
contentHeight: transformBoxColumn.height
contentWidth: transformBoxColumn.width
```

Also the `StackLayout` doesn't check if `transform_type` is "Rotate" in order to make the switch. It just assumes that a value that isn't "Translate" must be "Rotate." Perhaps not the best way to implement this but I couldn't see another way...

```qml
currentIndex: transform_type == "Translate" ? 0 : 1
Layout.preferredHeight: currentIndex == 0 ? translatePane.implicitHeight : rotatePane.implicitHeight
```

### Acceptance Criteria 

Add multiple different transformation types to a component and make sure the switch works correctly. Also make sure that transformations get a slightly smaller box compared to rotations.

### UI tests

No intentional changes to UI behavior.

### DIFF
--- a/resources/Qtmodels/TransformControls.qml
+++ b/resources/Qtmodels/TransformControls.qml
@@ -116,7 +116,7 @@ Item {
 
         Frame {
             id: transformBox
-            contentHeight: transformBoxStack.height + transformButtons.implicitHeight
+            contentHeight: transformBoxColumn.height
             contentWidth: Math.max(translatePane.implicitWidth, rotatePane.implicitWidth, transformButtons.implicitWidth)
             anchors.right: parent.right
             anchors.left: parent.left
--- a/resources/Qtmodels/TransformControls.qml
+++ b/resources/Qtmodels/TransformControls.qml
@@ -117,7 +117,6 @@ Item {
         Frame {
             id: transformBox
             contentHeight: transformBoxColumn.height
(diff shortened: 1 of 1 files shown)


rerpha:
changes look good but i get this error: `file:///home/jack/code/repos/nexus-constructor/resources/Qtmodels/TransformControls.qml: ReferenceError: parents is not defined`

looks like its coming from line 132 of TransformControls.qml

PR #286, opened 2019-05-17
234: Red cross for repeated component names

### LINKED ISSUE #234: Red cross for repeated component names
Using `InvalidInputCross` for the places where "Component names must be unique" appears

DolicaAkelloEgwel:
### Issue

Closes #234 

### Description of work

Fixed an issue that caused the bad units mouse-over message to be accessible even when the cross isn't visible. Added `InvalidInputCross` in places where it's possible to enter a component name which required adding a success signal to the `NameValidator`.

### Acceptance Criteria 

Check that mouse-over message can't be accessed when the cross is invisible. Try to give a component a repeated name and check that the cross appears/disappears when it should. Also go through UI tests.

### UI tests

Added a section about [validating component names](https://github.com/ess-dmsc/nexus-constructor/blob/234_Red_cross_for_repeated_component_names/tests/UITests.md#validating-component-names). Also removed the old test for this as it's out-of-date.

### DIFF
--- a/resources/Qtmodels/EditComponentWindow.qml
+++ b/resources/Qtmodels/EditComponentWindow.qml
@@ -51,7 +51,7 @@ ExpandingWindow {
                 columns: 3
 
                 Label {
-                    text: "Name: "
+                    text: "Name : "
                 }
                 TextField {
                     id: nameField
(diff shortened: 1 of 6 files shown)


resources/Qtmodels/EditComponentWindow.qml
```diff
@@ -79,12 +87,12 @@ ExpandingWindow {
                 Label {
                     id: transformLabel
                     text: "Transform:"
-                    Layout.columnSpan: 2
+                    Layout.columnSpan: 3
```
matthew-d-jones: As you want it to span all columns perhaps it is better to use `parent.columns` rather than hardcoding as `3`?

PR #307, opened 2019-06-05
Add component dialog

### LINKED ISSUE #110: Detailed editing of entire NeXus Structure
I don't see any way around having an editable treeview for the entire NeXus structure, but other suggestions are welcome.
A tab setup to switch between the basic component list and a full treeview seems like a nice idea.
This ticket needs to be discussed and planned in more detail before starting any coding!

rerpha:
### Issue

Closes #303 
Closes #110

### Description of work

Added an "add component" dialog. 

This should work and add components into the NeXus file as well as the component model. Transformations are not in this window. 

### Acceptance Criteria 

Add a few components and export to nexus file, then check the components have been added in the nexus file. There isn't really a way of checking the instrument model easily without going in with a debugger but to test you could put a breakpoint on somewhere after it has added the component and check it exists in the instrument model. 

### UI tests

N/A


matthew-d-jones:
Note: one needs to `git submodule update --init` for this branch to work.

nexus_constructor/addcomponentwindow.py
```diff
@@ -0,0 +1,253 @@
+from enum import Enum
```
matthew-d-jones: Can we use snake_case for file names?

nexus_constructor/addcomponentwindow.py
```diff
@@ -0,0 +1,253 @@
...
+        self.component_types = make_dictionary_of_class_definitions(
+            os.path.abspath(os.path.join(os.curdir, "definitions"))
+        )
+
+    def setupUi(self, parent_dialog):
+        """ Sets up push buttons and validators for the add component window. """
+        super().setupUi(parent_dialog)
+
+        # Connect the button calls with functions
+        self.buttonBox.button(QDialogButtonBox.Ok).clicked.connect(self.on_ok)
```
matthew-d-jones: Are these camelCase names created by NeXus Designer, or can we change them to snake_case?

nexus_constructor/component_type.py
```diff
@@ -14,3 +17,37 @@ class ComponentType(Enum):
...
+
+
+def __list_base_class_files(repo_directory):
+    base_class_dir = os.path.join(repo_directory, "base_classes")
+    for file in os.listdir(base_class_dir):
+        if file.endswith(".nxdl.xml"):
+            yield os.path.join(base_class_dir, file)
+
+
+def make_dictionary_of_class_definitions(
```
matthew-d-jones: Could this be unit tested, or refactored to make unit testable?

matthew-d-jones:
"Pixel options" should only be displayed for NXmonitor, NXdetector and NXdetector_module.
The default vector for cylinder should be x=0, y=0, z=1

nexus_constructor/addcomponentwindow.py
rerpha: They're created by Qt Designer, but I think we can change them to snake_case. I think Qt UI files are by default camelCase, not sure if there's a standard.

nexus_constructor/add_component_window.py
```diff
@@ -0,0 +1,258 @@
...
+from nexus_constructor.validators import UnitValidator, NameValidator
+from nexus_constructor.nexus_wrapper import NexusWrapper
+from nexus_constructor.utils import file_dialog, validate_line_edit
+import os
+from functools import partial
+
+GEOMETRY_FILE_TYPES = {"OFF Files": ["off", "OFF"], "STL Files": ["stl", "STL"]}
+
+
+class FileValidator(QValidator):
```
matthew-d-jones: I suggest naming it more specifically `GeometryFileValidator` and moving it to its own file.

nexus_constructor/add_component_window.py
```diff
@@ -0,0 +1,258 @@
...
+        if self.meshRadioButton.isChecked():
+            geometry_model = OFFModel()
+            geometry_model.set_units(self.unitsLineEdit.text())
+            geometry_model.set_file(self.geometry_file_name)
+        else:
+            geometry_model = NoShapeModel()
+        return geometry_model
+
+    def on_ok(self):
+        component_type = self.componentTypeComboBox.currentText()
```
matthew-d-jones: Does "type" here mean "Nx_class"? If so let's use the terminology `nx_class` to reduce ambiguity.

nexus_constructor/component.py
```diff
@@ -1,74 +1,19 @@
...
-        ),
-    )
 
 
 @attr.s
 class Component:
     """Components of an instrument"""
 
-    component_type = attr.ib(ComponentType)
+    nx_class = attr.ib(str)
```
matthew-d-jones: Good, I think there are other places that would benefit from this change too.

nexus_constructor/component_type.py
```diff
@@ -1,16 +1,46 @@
...
+    class_definitions[nx_class_name] = []
+    try:
+        fields = xml_definition["field"]
+        try:
+            for field in fields:
+                class_definitions[nx_class_name].append(field["@name"])
+        except Exception:
+            class_definitions[nx_class_name].append(fields["@name"])
+    except KeyError:
+        # TODO: No key called "field"
```
matthew-d-jones: There is nothing "TODO" here, if there is no key called "field" then that is fine.

nexus_constructor/validators.py
```diff
@@ -32,27 +32,25 @@ def validate(self, input: str, pos: int):
...
             return QValidator.Intermediate
 
-        self.validationSuccess.emit()
+        self.isValid.emit(True)
         return QValidator.Acceptable
 
-    # Create signals because the QML LabeledTextField doesn't have "onAccepted"
-    validationSuccess = Signal()
-    validationFailed = Signal()
+    isValid = Signal(bool)
```
matthew-d-jones: Is this a standard signal name which must be camelCase, or can we use snake_case?

rerpha: we can use snake_case here i believe.

Issue #297, opened 2019-05-30
Widgets: get 3d view working 


rerpha:
Branch this feature off of `switch_to_widgets` and create pull requests to merge against this branch. 

There is a blank widget in place on `switch_to_widgets` but this is currently doing nothing. 
 
The logic can be pretty much copied across from what we're doing currently in QML but using qt3d instead.

matthew-d-jones:
Closed by #308

PR #308, opened 2019-06-06
297: Basic InstrumentView

DolicaAkelloEgwel:
### Issue

Makes some progress on #297 but doesn't close it.

### Description of work

Adds an instrument view with a sample cube, a beam, and some neutron animations. The sample cube is created by adding a cuboid shape in `InstrumentView` instead of reading from an OFF file like before. The axis indicator is currently in a separate branch.

This makes the sample appear shaded in the place where it overlaps with the beam. The QML didn't do this but I think it makes more sense this way.

Also zooming so that the camera is inside the beam/sample makes it "disappear" but this doesn't happen with QML. Not sure if that's considered serious or not?

There is now a guide for adding custom widgets to Qt Designer [here](https://github.com/ess-dmsc/nexus-constructor/wiki/Qt-Designer-Guide#using-custom-widgets).

### Acceptance Criteria 

Compare it to the QML instrument view and check that the appearance/behavior is mostly the same.

### UI tests

n/a


ui/InstrumentView.py
```diff
@@ -0,0 +1,146 @@
+from PySide2.QtWidgets import QWidget, QVBoxLayout
```
rerpha: I think only the generated UI files and resources like icons etc. should be in this file. Just my opinion, but i think we should move these out into the `nexus_constructor/` directory?

ui/mainwindow.py
```diff
@@ -48,14 +49,19 @@ def setupUi(self, MainWindow):
...
         sizePolicy.setVerticalStretch(0)
         sizePolicy.setHeightForWidth(self.sceneWidget.sizePolicy().hasHeightForWidth())
         self.sceneWidget.setSizePolicy(sizePolicy)
         self.sceneWidget.setMinimumSize(QtCore.QSize(745, 0))
         self.sceneWidget.setObjectName("sceneWidget")
-        self.gridLayout_3.addWidget(self.sceneWidget, 0, 1, 1, 2)
+        self.instrumentView.setSizePolicy(sizePolicy)
+        self.instrumentView.setMinimumSize(QtCore.QSize(745, 0))
+        self.instrumentView.setObjectName("instrumentView")
+        # self.gridLayout_3.addWidget(self.sceneWidget, 0, 1, 1, 2)
```
rerpha: you need to make these changes in mainwindow.py in the `nexus_constructor/` directory, these will be removed when the ui file gets changed and `pyside2-uic` is used. The one in `nexus_constructor/` is the actual implementation.

ui/mainwindow.ui
```diff
@@ -106,20 +106,28 @@
...
   <action name="actionExport_to_NeXus_file">
    <property name="text">
-    <string>Export to NeXus file</string>
+    <string>&amp;Export to NeXus file</string>
    </property>
   </action>
   <action name="actionExport_to_Filewriter_JSON">
    <property name="text">
-    <string>Export to Filewriter JSON</string>
+    <string>Export to &amp;Filewriter JSON</string>
```
rerpha: something has gone a bit weird here and various other places in the UI file

ui/mainwindow.ui
```diff
@@ -89,13 +89,13 @@
     <rect>
      <x>0</x>
      <y>0</y>
-     <width>1263</width>
-     <height>19</height>
+     <width>1271</width>
+     <height>20</height>
```
rerpha: do we need to change this? all it affects is the initial size of the window?

ui/mainwindow.ui
```diff
@@ -120,6 +120,14 @@
    </property>
   </action>
  </widget>
+ <customwidgets>
+  <customwidget>
+   <class>InstrumentView</class>
+   <extends>QWidget</extends>
+   <header>nexus_constructor/InstrumentView.h</header>
```
rerpha: Could you stick something in the WIKI about how this works? cheers

ui/mainwindow.py
```diff
@@ -87,8 +87,9 @@ def retranslateUi(self, MainWindow):
         self.pushButton.setText(QtWidgets.QApplication.translate("MainWindow", "Add component", None, -1))
         self.tabWidget.setTabText(self.tabWidget.indexOf(self.tab_2), QtWidgets.QApplication.translate("MainWindow", "Components", None, -1))
         self.tabWidget.setTabText(self.tabWidget.indexOf(self.tab), QtWidgets.QApplication.translate("MainWindow", "NeXus File Layout", None, -1))
-        self.menuFile.setTitle(QtWidgets.QApplication.translate("MainWindow", "File", None, -1))
-        self.actionOpen_NeXus_file.setText(QtWidgets.QApplication.translate("MainWindow", "Open NeXus file", None, -1))
-        self.actionExport_to_NeXus_file.setText(QtWidgets.QApplication.translate("MainWindow", "Export to NeXus file", None, -1))
-        self.actionExport_to_Filewriter_JSON.setText(QtWidgets.QApplication.translate("MainWindow", "Export to Filewriter JSON", None, -1))
+        self.menuFile.setTitle(QtWidgets.QApplication.translate("MainWindow", "Fi&le", None, -1))
```
rerpha: still some ampersands here

ui/mainwindow.ui
DolicaAkelloEgwel: I've tried to do that here https://github.com/ess-dmsc/nexus-constructor/wiki/Qt-Designer-Guide#using-custom-widgets

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
+from PySide2.QtWidgets import QWidget, QVBoxLayout
+from PySide2.Qt3DExtras import Qt3DExtras
+from PySide2.Qt3DCore import Qt3DCore
+from PySide2.QtCore import QPropertyAnimation
+from PySide2.QtGui import QVector3D, QColor, QMatrix4x4
+from nexus_constructor.NeutronAnimationController import NeutronAnimationController
+
+
+class InstrumentView(QWidget):
```
rerpha: Perhaps just add a docstring here describing the overall purpose of the class, and same with `NeutronAnimationController`.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
...
+        Creates the materials for the the objects that will inhabit the instrument view.
+        """
+        red = QColor("red")
+        black = QColor("black")
+        grey = QColor("grey")
+        blue = QColor("blue")
+        light_blue = QColor("lightblue")
+        dark_red = QColor("#b00")
+
+        self.grey_material.setAmbient(black)
```
rerpha: looks like the following lines are almost duplication, could refactor this out into a separate function and test that it applies the ambient and diffuse values.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
...
+
+            neutron_transform = Qt3DCore.QTransform()
+            neutron_animation_controller = NeutronAnimationController(
+                x_offsets[i], y_offsets[i], neutron_transform
+            )
+            neutron_animation_controller.set_target(neutron_transform)
+
+            # Instruct the NeutronAnimationController to move the neutron along the z-axis from -40 to 0
+            neutron_animation = QPropertyAnimation(neutron_transform)
+            neutron_animation.setTargetObject(neutron_animation_controller)
```
rerpha: this could be split up and tested, pass in the QPropertyAnimation into the refactored function and test these properties are set on it.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
...
+            neutron_animation.setPropertyName(b"distance")
+            neutron_animation.setStartValue(-self.cylinder_length)
+            neutron_animation.setEndValue(0)
+            neutron_animation.setDuration(500 + time_span_offsets[i])
+            neutron_animation.setLoopCount(-1)
+            neutron_animation.start()
+
+            self.neutron_entities.append(neutron_entity)
+            self.neutron_meshes.append(neutron_mesh)
+            self.neutron_transforms.append(neutron_transform)
```
rerpha: Add a comment above these to mention that this is actually a workaround because there is a bug in the PySide2 bindings that requires us to keep these Qt3D entities in scope so they are not destroyed by C++. 

I wonder here if a dict might be better, as we are just using it to keep these variables in scope. It could be laid out like this for example: 
```python
entity_store = {
"neutron_entities" : [],
"neutron_meshes" : []
}
``` 
etc. 

this way when they fix this bug we can just remove the one dictionary and it's calls.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
...
+        self.beam_material.setAlpha(0.5)
+
+        self.green_material.setAmbient(grey)
+        self.green_material.setDiffuse(grey)
+
+    def create_sample_cube(self):
+        """
+        Creates the initial sample cube.
+        """
+        self.cube_mesh.setXExtent(1)
```
rerpha: same here with refactoring. pass in the cube object into this function, make it a free function, and check that these properties are set on it.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,157 @@
...
+        self.cube_mesh.setZExtent(1)
+
+        self.cube_entity.addComponent(self.cube_mesh)
+        self.cube_entity.addComponent(self.red_material)
+
+    def create_beam_cylinder(self):
+        """
+        Creates the initial beam cylinder.
+        """
+        self.cylinder_mesh.setRadius(2.5)
```
rerpha: As above.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,260 @@
...
+        # Initialise cube objects
+        self.sample_cube_dimensions = [1, 1, 1]
+        self.cube_entity = Qt3DCore.QEntity(self.root_entity)
+        self.cube_mesh = Qt3DExtras.QCuboidMesh()
+
+        self.num_neutrons = 9
+
+        # Create lists for neutron-related objects
+        self.neutron_entities = []
+        self.neutron_meshes = []
```
rerpha: let's make this a dictionary, it's not used for anything other than keeping variables in scope.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,260 @@
...
+            self.cylinder_mesh, 2.5, self.cylinder_length, 2
+        )
+        self.set_beam_transform(self.cylinder_transform)
+        self.add_components_to_entity(
+            self.cylinder_entity,
+            [self.cylinder_mesh, self.beam_material, self.cylinder_transform],
+        )
+
+    @staticmethod
+    def set_sphere_mesh_radius(sphere_mesh, radius):
```
rerpha: pass in a mock to these tests and check that `setRadius` has been called here for example.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,260 @@
...
+        neutron_animation_controller = NeutronAnimationController(
+            x_offset, y_offset, neutron_transform
+        )
+        neutron_animation_controller.set_target(neutron_transform)
+
+        return neutron_animation_controller
+
+    @staticmethod
+    def create_neutron_animation(
+        neutron_transform,
```
rerpha: docstrings on all methods describing what they do and what the parameters are

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def __init__(self, parent):
+        super().__init__()
+        lay = QVBoxLayout(self)
+        self.view = Qt3DExtras.Qt3DWindow()
+        self.view.defaultFrameGraph().setClearColor(QColor("lightgrey"))
+        container = QWidget.createWindowContainer(self.view)
+        lay.addWidget(container)
+
+        self.view.camera().lens().setPerspectiveProjection(45, 16 / 9, 0.1, 1000)
```
rerpha: maybe comments here describing why we have these magic numbers that set the position etc. 

explain why they're set to what they are

nexus_constructor/InstrumentView.py
```diff
@@ -47,17 +47,18 @@ def __init__(self, parent):
         self.num_neutrons = 9
 
         # Create lists for neutron-related objects
-        self.neutron_entities = []
-        self.neutron_meshes = []
-        self.neutron_transforms = []
+        self.neutron_objects = {
```
rerpha: stick a comment here that they're just stored because pyside2 qt3d has a bug that requires variables to be kept in scope

nexus_constructor/InstrumentView.py
DolicaAkelloEgwel: I'll add something although this was copied from QML so I don't entirely understand it myself.

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,273 @@
...
+        # Initialise cube objects
+        self.sample_cube_dimensions = [1, 1, 1]
+        self.cube_entity = Qt3DCore.QEntity(self.root_entity)
+        self.cube_mesh = Qt3DExtras.QCuboidMesh()
+
+        self.num_neutrons = 9
+
+        """
+        Create a dictionary for neutron-related objects so that they are always in scope and not destroyed by C++
+        """
```
matthew-d-jones: Should be a line comment not a block comment

nexus_constructor/InstrumentView.py
```diff
@@ -0,0 +1,273 @@
...
+    @staticmethod
+    def set_beam_transform(cylinder_transform):
+        """
+        Configures the transform for the beam cylinder by giving it a matrix.
+        :param cylinder_transform: A QTransform object.
+        """
+        """
+        This matrix will turn the cylinder sideways and then move it "backwards" in the z-direction by 20 units so that
+        it ends at the location of the sample.
+        """
```
matthew-d-jones: Please combine these to a single block comment

PR #319, opened 2019-06-13
312: Fix Executables and Reduce Size

### LINKED ISSUE #312: Widgets: fix executables and reduce size
If the executables from jenkins don't work we need to fix them.

We should also exclude module components that we're not using in the setup.py script to reduce the size of the executables such as: 

- `/lib/PySide2/Qt/qml/`
- `/definitions/manual` 

etc

DolicaAkelloEgwel:
### Issue

Closes #312

### Description of work

Instructs setup.py to remove folders that aren't required for the executable to run.

### Acceptance Criteria 

Run `python setup.py build_exe` and check that the executable works, file reading/writing works, etc.

### UI tests

n/a


matthew-d-jones:
@DolicaAkelloEgwel We should only try to exclude a few of the large volume directories which are unneeded. If we try to exclude every little unused file individually then this will be unmaintainable, these will change when we change or update dependencies.

setup.py
```diff
@@ -40,6 +40,94 @@
...
+    "lib/distutils",
+    "lib/mock",
+    "lib/PyQt5",
+    "lib/babel",
+    "lib/sphinx",
+    "platforms",
+    "imageformats",
+    "definitions",
+    "mpl-data",
+    "lib/PySide2/Qt/qml",
```
rerpha: this will only work on linux as it contains forward slashes. Windows won't work properly here. Use os.path.join instead of the slashes like so: 

```python 
os.path.join(os.path.join(os.path.join("lib", "PySide2"), "Qt"), "qml")
```

I'm not sure if there's a nicer way to do this, but it reads OK.

setup.py
```diff
@@ -64,7 +64,7 @@
 
 larger_folders = [
     "platforms",
-    "definitions/manual",
+    os.path.join("definitions", "qml"),
```
rerpha: need "definitions" and "manual" here instead of "qml"

setup.py
```diff
@@ -40,6 +62,14 @@
     removable = []
     extension = ""
 
+larger_folders = [
+    "platforms",
+    os.path.join("definitions", "manual"),
+    "imageformats",
+    "mpl-data",
+    os.path.join(os.path.join(os.path.join("lib", "PySide2"), "Qt"), "qml"),
```
matthew-d-jones: No need to need these, you can just do
```
os.path.join("lib", "PySide2", "Qt", "qml")
```

PR #324, opened 2019-06-17
317: 3D view should use the components model to display 3D models

### LINKED ISSUE #317: Widgets: 3d view should use the components model to display 3d models
Currently shows the animations but not the components in `instrumentmodel`

we can probably use the logic that we were using in QML to do this.

DolicaAkelloEgwel:
### Issue

Closes #317

### Description of work

Allows "Add Component" to add different components types to the 3D view. At the moment there is no way of deleting components/transformations or adding transformations. I have implemented a method for deleting components that worked when I tried it out. Once the button for deleting components gets added that should be simple to connect to the `delete_component` method.

### Acceptance Criteria 

Add components with different geometry types (none, cylinder, mesh) and check that they appear in the instrument view. Also check that changing the units gives the correct result.

### UI tests

n/a

### DIFF
--- a/nexus_constructor/instrument_view.py
+++ b/nexus_constructor/instrument_view.py
@@ -78,6 +78,7 @@ class InstrumentView(QWidget):
 
         self.component_meshes = {}
         self.component_entities = {}
+        self.component_translations = {}
 
     def add_component(self, name, geometry):
         """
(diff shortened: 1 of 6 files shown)


nexus_constructor/instrument_view.py
```diff
@@ -93,6 +94,24 @@ def add_component(self, name, geometry):
...
+    def delete_component(self, name):
+        """
+        Delete a component from the InstrumentView by removing the components and entities from the dictionaries.
+        :param name: The name of the component.
+        """
+        try:
+            del self.component_entities[name]
+            del self.component_meshes[name]
+            del self.component_translations[name]
+        except KeyError:
```
rerpha: stick a print statement here saying something like "unable to delete component" - just avoids other people having to guess where the code is failing in the future rather than failing silently

Issue #334, opened 2019-06-24
Widgets: add geometry information to component group in nexus file


rerpha:
Replace qml geometry and data models, should allow also removing the old `Component` class (which is replaced by `ComponentModel`). Look at `ComponentModel` and `TransformationModel` for example of how to replace the qml model with a wrapper over the NeXus file.

PR #337, opened 2019-06-25
Refactor component and transform

### LINKED ISSUE #311: Widgets: stop taking ownership of list for list view, and use nexus entirely for model
Currently the `NeXusWrapper` has ownership of an `InstrumentModel` which is essentially just a Qt list that contains all the components. My suggestion is that we don't take ownership of this, and instead generate it when needs be from the Nexus file itself. 

there is already a `get_components_list` function in `NexusWrapper` so the interface will not have to be changed even if people are working on stuff that relies on the component list so long as they use this method to access it.

### LINKED ISSUE #336: Refactor component and transformation
These should provide models by simply being an interface to the data in the nexus file, via `NexusWrapper`.

As previously discussed we will keep a tree (connected acyclic graph) structure of transformations in order to handle "depends_on" chains and avoid deleting transformations which have dependent transformations.

matthew-d-jones:
### Issue

Closes #336
Closes #311

### Description of work

This is intended to demonstrate my proposal for what to do with `Component`. Please take a look and see if you think this approach makes sense.

`Component` is the existing implementation and `ComponentModel` is the new implementation. We can rename `ComponentModel` to `Component` once the old implementation is not used anywhere.
`Transformation` would need to be updated in a similar way to `Component` to just wrap the data in the NeXus file.

This is WIP, more needs to be implemented and including some unit tests. However it does run in its current state, and the "Add Component" interface works.

For @SkyToGround's component view:
`Instrument.get_component_list()` gives you the component list. `ComponentModel.transforms` gives you the list of transformations for a component.


nexus_constructor/component.py
```diff
@@ -1,13 +1,83 @@
...
+    def description(self):
+        return self.file.get_field_value(self.group, "description")
+
+    @description.setter
+    def description(self, description: str):
+        if description:
+            self.file.set_field_value(self.group, "description", description, str)
+
+    @property
+    def transforms(self):
```
SkyToGround: For creating the UI, making the list of transformations hold a pointer to the component makes things a lot easier for me. See the [simple implementation](https://github.com/ess-dmsc/nexus-constructor/blob/add_component_view/nexus_constructor/component.py) that I have used for reference.

nexus_constructor/component.py
```diff
@@ -1,13 +1,83 @@
...
+        """
+        Gets all transforms in the depends_on chain for this component
+        :return: List of transforms
+        """
+        transforms = []
```
SkyToGround: It would be nice if the list is updated with new transformations instead of being re-created every time you call `transforms()`. With that said, it is probably not that hard to work around this.

nexus_constructor/instrument.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def remove_component(self, component: ComponentModel):
+        """
+        Removes a component group from the NeXus file
+        :param component: The component to be removed
+        """
+        self.nexus.delete_group(component.group)
+
+    def get_component_list(self):
+        component_list = []
```
SkyToGround: Similar to previous comment, it would be nice if this list is updated instead of re-created.

nexus_constructor/component.py
matthew-d-jones: To reduce complexity I'm trying to avoid that there is any cached data layer between the in-memory NeXus file and the GUI.

nexus_constructor/component.py
matthew-d-jones: We also need to figure out what transforms your interface should actually display. The list returned here is the full depends_on chain for this component. Only some, or none, of those transformations may live in the component group in the file.

nexus_constructor/component.py
```diff
@@ -96,7 +100,7 @@ def transforms_full_chain(self):
         Gets all transforms in the depends_on chain for this component
         :return: List of transforms
         """
-        transforms = []
+        transforms = TransformationsList(self)
```
SkyToGround: Thanks!

matthew-d-jones: Just patching a few holes in the test coverage then I'll promote this PR for review

matthew-d-jones:
Hopefully these changes are enough for Jonas to be able to progress and to clarify how we should proceed with replacing the other qml data models. The old `Component` still exists until we clean up geometry models in #334 .

matthew-d-jones:
@SkyToGround  `Component.transforms` gives you the transforms which should be displayed under the component in your treeview, as we discussed yesterday. However it does not return the "link to external transforms".

SkyToGround:
Thanks! I will try to have a look at it during the weekend. (I just realised how close to September we are).

SkyToGround:
@rerpha fair enough. Go ahead! :-)

matthew-d-jones:
@SkyToGround We'll wait on your review before merging as it directly affects what you are working on, and it isn't blocking Jack from sorting out the JSON writing.

nexus_constructor/component.py
```diff
@@ -1,13 +1,242 @@
...
+    ):
+        """
+        Recursive function, appends each transform in depends_on chain to transforms list
+        :param depends_on: The next depends_on string to find the next transformation in the chain
+        :param transforms: The list to populate with transformations
+        :param local_only: If True then only add transformations which are stored within this component
+        """
+        if depends_on and depends_on != ".":
+            transform_dataset = self.file.nexus_file[depends_on]
+            if local_only and transform_dataset.parent.parent.name != self.group.name:
```
SkyToGround: 👍

PR #340, opened 2019-06-28
Remove component filter - only used by QML

rerpha:
### Issue

None

### Description of work

Removed the model filters, only used by QML so no longer needed. 

### Acceptance Criteria 

No functional changes. Just check that nothing has broken (it shouldn't have, because this code is not used or tested) 

### UI tests

N/A


matthew-d-jones:
Already done in #337

PR #349, opened 2019-06-29
Transforms keep track of their dependents

### LINKED ISSUE #157: Transform objects should know their parents and children
Currently the information about the order that Transforms are performed is contained within the Transform_Model. There are a number of issues with this:
* It is unclear on the relationship between components and transform models
* The logic inside transform_model is complicated

It was proposed during #142 that the logic of parent/child transforms should instead be contained within the transformation objects themselves. e.g. each will hold a reference to it's parent and children. Components can then hold references to transformations and there will not be a need for a class dedicated to knowing about order.

Edge cases that should be thought about with this are:
* Transformations do not need to be within a component
* What happens when a transform is removed? Are child transforms rerouted or do we just not allow this?

This ticket is to:
* Better understand the logic within transform_model
* Confirm that the proposed data structure better suits the problem and covers all use cases
* Implement the solution
* Add unit tests where appropriate
* Document the data structure

matthew-d-jones:
### Issue

Closes #157

### Description of work

Keep track of what transforms and components depend on a given transform by storing this information in an `dependee_of` attribute in the transformation. This is then used to stop a transformation being deleted if anything else is still dependent on it. This is covered by unit tests in `test_component.py`.
As `dependee_of` is not part of the NeXus standard we cannot rely on it being present in any existing NeXus files which we load, so I make sure it is generated on file load. This is covered by a unit test in `test_instrument.py`.


matthew-d-jones:
Recreating PR as the diff is not updating

PR #352, opened 2019-07-01
Adding nexus file to JSON functionality

### LINKED ISSUE #341: Widgets: write filewriter json from nexus file rather than instrumentmodel
We need to generate the filewriter JSON commands from the nexus file itself rather than the current method of using the contents of `InstrumentModel`. As this class is being deprecated and removed, we should use the nexus file as the starting point to generate the JSON instead. 

Matt potentially has some code that will generate the fw JSON from the in-memory nexus file.

rerpha:
### Issue

Closes #341 

### Description of work

- Copied code from ess-dmsc/nexus-to-json and modified to use h5py rather than nexusformat module. 
- Wrote tests for the `NexusToDictConverter` class
- Tied new functionality into UI 

### Acceptance Criteria 

- Reviewer is happy with unit tests 

- Test adding a few components: 
  - test the JSON is as expected 
  - run through the filewriter to see if it complains about bad commands

### UI tests

N/A

### DIFF
--- a/nexus_constructor/main_window.py
+++ b/nexus_constructor/main_window.py
@@ -78,10 +78,9 @@ class MainWindow(Ui_MainWindow):
 
     def save_to_filewriter_json(self):
         filename = file_dialog(True, "Save JSON File", JSON_FILE_TYPES)
-        self.instrument.nexus.save_file(filename)
         if filename:
             with open(filename, "w") as file:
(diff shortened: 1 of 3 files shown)


rerpha:
Marking as in progress so i can add some more tests to get the coverage to as close to 100% as i can

rerpha:
Filewriter doesn't complain and handles the command :+1:

PR #354, opened 2019-07-01
Add suggestion when component name fails to validate 

### LINKED ISSUE #320: Widgets: Automatically populate Name field of Add Component interface
Same functionality as we had before the switch to widgets, suggest name "detector_1", or if that already exists then "detector_2" etc.

Can use `generate_unique_name()` in `ui_utils.py`, please also add unit tests for this method.

rerpha:
### Issue

Closes #320 

### Description of work

Adds a suggestion_callable parameter to `validate_line_edit` which gives a suggestion based on the `nx_class` selected and the existing names in the component list. 

Suggestions are in the tooltip that's shown when hovering over the name line edit.

### Acceptance Criteria 

- Test with no components - anything other than a blank name should be valid, but it should still give you a suggestion. 

- Test with components that already exist - should give a suggestion. 

### UI tests

N/A

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -103,7 +103,7 @@ def test_GIVEN_name_with_1_in_already_WHEN_generating_unique_name_THEN_number_is
     )
 
 
-def test_GIVEN_name_with_1_as_prefix_and_name_with_11_already_in_list_WHEN_generating_unique_name_THEN_following_number_is_incremented_and_name_is_unique():
+def test_GIVEN_name_with_1_as_suffix_and_name_with_11_already_in_list_WHEN_generating_unique_name_THEN_following_number_is_incremented_and_name_is_unique():
     comp = DummyComponent("something1")
 
     assert (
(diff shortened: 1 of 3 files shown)


rerpha:
marking this as in progress, have just realised there are no tests for `generate_unique_name()` so i am going to write some for it.

tests/test_utils.py
```diff
@@ -45,3 +47,65 @@ def test_GIVEN_invalid_WHEN_validating_line_edit_with_no_tooltip_THEN_tooltip_is
...
+
+    assert (
+        generate_unique_name(
+            comp.name, [DummyComponent("something1"), DummyComponent("something2")]
+        )
+        == comp.name + "1"
+    )
+
+
+def test_GIVEN_name_with_1_as_prefix_and_name_with_11_already_in_list_WHEN_generating_unique_name_THEN_following_number_is_incremented_and_name_is_unique():
```
matthew-d-jones: `prefix` -> `suffix`

PR #356, opened 2019-07-02
346: Use neutron animation in the gnomon

### LINKED ISSUE #346: Use the neutron animation in the gnomon
I wonder if we could replace the z axis cylinder in the gnomon with the neutron beam animation, and remove the animation from the main view?
It would be neat to show the beam direction in that way in the gnomon and would solve the problem of the animation being "in the way" in the main view.

DolicaAkelloEgwel:
### Issue

Closes #346

### Description of work

Moves the beam cylinder and neutron animation to the gnomon. Methods related to the neutrons and beam were moved to gnomon.py. Tests for these methods were also moved to test_gnomon.py. I shrunk the size of the neutrons and beam in order to make this look better. Also had the x/y offsets halved for the `NeutronAnimationController`.

There is also a bit of a strange effect when the axis labels are in front of the neutrons/cylinder. Not sure how important that is.

### Acceptance Criteria 

Check that all the tests pass. Also check that the gnomon still fits entirely inside its viewport box even if you move it around.

### UI tests

n/a

### DIFF
--- a/nexus_constructor/gnomon.py
+++ b/nexus_constructor/gnomon.py
@@ -89,6 +89,8 @@ class Gnomon:
 
         self.num_neutrons = 9
 
+        self.neutron_animation_length = self.gnomon_cylinder_length * 1.5
+
         # Create a dictionary for neutron-related objects so that they are always in scope and not destroyed by C++
         self.neutron_objects = {
             "entities": [],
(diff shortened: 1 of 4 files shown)


matthew-d-jones:
@rerpha I think it would be good to have axes in the main view, but I'll merge this and make a new ticket.

PR #357, opened 2019-07-02
Adding dialog for when a nexus file has multiple entries 

### LINKED ISSUE #348: Incorrect entry and instrument when loading existing NeXus file
nexus_wrapper has an `open_file` method to load an existing NeXus file, however it does not set `self.entry` and `self.instrument` to be consistent with this file. Therefore if they are called anything other than "entry" and "instrument" the application will not function correctly.

Need to consider what to do if there are multiple entry groups or instruments in the file.
I suggest:
- allow user to pick which entry group they want to use
- throw error if there are multiple `NXinstrument` groups

rerpha:
### Issue

Closes #348 

### Description of work
Adds a dialog that allows the user to choose between `NXentry`s if there are multiple in a file. 

- Added signal for when there are multiple entries in a file
- Added dialog with combo box that shows the names of the entries in a file
- Added function for finding the first `NXinstrument` group in an entry group 

### Acceptance Criteria 

Unit tests should cover all of the new added functionality minus the UI parts until we can get those working. 

Should load nexus files as before but if there are multiple entries it should show the dialog and when selected should select the entry and instrument correctly. to test this add a new component then check the tree view. 

### UI tests

n/a

### DIFF
--- a/nexus_constructor/main_window.py
+++ b/nexus_constructor/main_window.py
@@ -50,18 +50,26 @@ class MainWindow(Ui_MainWindow):
         self.widget.setVisible(True)
 
     def show_entries_dialog(self, map_of_entries: dict, nexus_file):
+        """
+        Shows the entries dialog when loading a nexus file if there are multiple entries.
+        :param map_of_entries: A map of the entry groups, with the key being the name of the group and value being the actual h5py group object.
+        :param nexus_file: A reference to the nexus file.
+        """
         self.entries_dialog = QDialog()
         self.entries_dialog.setMinimumWidth(400)
         self.entries_dialog.setWindowTitle(
             "Multiple Entries found. Please choose the entry name from the list."
         )
         combo = QComboBox()
+
+        # Populate the combo box with the names of the entry groups.
         [combo.addItem(x) for x in map_of_entries.keys()]
 
         ok_button = QPushButton()
         ok_button.setText("OK")
         ok_button.clicked.connect(self.entries_dialog.close)
 
+        # Connect the clicked signal of the ok_button to instrument.load_file and pass the file and entry group object.
         ok_button.clicked.connect(
             partial(
                 self.instrument.nexus.load_file,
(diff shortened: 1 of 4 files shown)


nexus_constructor/nexus/nexus_wrapper.py
```diff
@@ -79,13 +80,44 @@ def open_file(self, filename):
...
                 filename, mode="r", backing_store=False, driver="core"
             )
-            self._load_file(nexus_file)
-            print("NeXus file loaded")
-            self._emit_file()
 
-    def _load_file(self, nexus_file: h5py.File):
+            self.find_entries_in_file(nexus_file)
+
+    def find_entries_in_file(self, nexus_file):
```
matthew-d-jones: missing type hint

nexus_constructor/main_window.py
```diff
@@ -39,14 +41,49 @@ def setupUi(self, main_window):
...
-        # self.listView.setModel(self.nexus_wrapper.get_component_list())
+        self.instrument.nexus.show_entries_dialog.connect(self.show_entries_dialog)
 
         self.instrument.nexus.component_added.connect(self.sceneWidget.add_component)
 
         self.set_up_warning_window()
 
         self.widget.setVisible(True)
 
+    def show_entries_dialog(self, map_of_entries: dict, nexus_file):
```
matthew-d-jones: missing type hint

tests/test_nexus_wrapper.py
```diff
@@ -24,3 +27,85 @@ def test_nxs_is_appended_to_filename():
...
+    wrapper = NexusWrapper(filename="test_nw5")
+
+    wrapper.find_entries_in_file(file)
+
+    assert wrapper.nexus_file == file
+    assert wrapper.entry == entry
+    assert wrapper.instrument == inst_group
+
+
+def test_GIVEN_multiple_entry_groups_in_file_WHEN_finding_entry_THEN_signal_is_emitted_with_correct_arguments():
```
matthew-d-jones: "entry_options" would be a little more elucidating than "correct_arguments"

PR #359, opened 2019-07-04
Adding scalar fields 

rerpha:
### Issue

None

### Description of work

Adds the ability to add field widgets in the add component window and adds scalar datasets to the nexus file. Does not add any other field type yet. 

- Added fields classes in `component_fields.py`
- Added some tests in `test_component_fields.py`
- Added some tests in `test_validators.py`

### Acceptance Criteria 

- Tests pass 
- Coverage will decrease slightly as there are some UI additions but should only be marginal 
- Adding a scalar field when adding a component should show the field in the nexus tree structure when added. 

### UI tests

N/A


matthew-d-jones:
Adding selection functionality to the fields to allow deletion is a good solution. However, the capturing the click event for that seems to mean i have to double click to use the combo boxes. The line edits only require one click though, so not sure exactly what is happening...

rerpha:
Marking as in progress whilst i sort out the validation of the value line edit.

rerpha:
thought whilst i was in `validators.py` i would just try and up the test coverage and test `GeometryFileValidator` so have added some tests for that and closed #367

nexus_constructor/add_component_window.py
```diff
@@ -122,6 +126,28 @@ def setupUi(self, parent_dialog):
...
+        item.setSizeHint(field.sizeHint())
+        self.fieldsListWidget.addItem(item)
+        self.fieldsListWidget.setItemWidget(item, field)
+
+    def select_field(self, widget):
+        self.selected_widget = widget
+        self.fieldsListWidget.setItemSelected(widget, True)
+
+    def remove_field(self):
+        self.fieldsListWidget.takeItem(self.fieldsListWidget.row(self.selected_widget))
```
DominicOram: If you use:
```python
selected_row = self.fieldsListWidget.row(self.fieldsListWidget.currentItem())
self.fieldsListWidget.takeItem(selected_row)
```
You will no longer need to hold `self.selected_widget` in this class

nexus_constructor/add_component_window.py
```diff
@@ -122,6 +126,28 @@ def setupUi(self, parent_dialog):
         # Validate the default values set by the UI
         self.unitsLineEdit.validator().validate(self.unitsLineEdit.text(), 0)
         self.nameLineEdit.validator().validate(self.nameLineEdit.text(), 0)
+        self.addFieldPushButton.clicked.connect(self.add_field)
+        self.removeFieldPushButton.clicked.connect(self.remove_field)
+
+        # Set whatever the default nx_class is so the fields autocompleter can use the possible fields in the nx_class
+        self.on_nx_class_changed()
+
+        self.fieldsListWidget.itemClicked.connect(self.select_field)
```
DominicOram: I think that if you implement the comment below to remove `self.selected_widget` this is no longer required.

nexus_constructor/component_fields.py
```diff
@@ -0,0 +1,190 @@
...
+                self.value_line_edit,
+                tooltip_on_accept="Value is cast-able to numpy type.",
+                tooltip_on_reject="Value is not cast-able to selected numpy type.",
+            )
+        )
+        self.value_line_edit.validator().validate(self.value_line_edit.text(), 0)
+
+        self.nx_class_combo = QComboBox()
+
+        self.edit_button = QPushButton("Edit")
```
DominicOram: I have to double click the edit button for the window to appear.

nexus_constructor/component_fields.py
```diff
@@ -0,0 +1,190 @@
...
+    @property
+    def dtype(self):
+        if self.field_type == FieldType.scalar_dataset:
+            return self.value.dtype
+        if self.field_type == FieldType.array_dataset:
+            return np.array(self.value).dtype
+
+    @property
+    def value(self):
+        if self.field_type == FieldType.scalar_dataset:
```
DominicOram: Having all these equality checks in this class is messy and it's only going to grow. I wonder if you could have a class that describes the part of the field that changes depending on it's type and have a inheritance structure that describes the different behaviour for each type? Or do you think that'll be messier?

nexus_constructor/add_component_window.py
rerpha: I have no idea why, but i can't get that to work. `currentRow()` or `currentItem()` seem to never change even if the item is selected.

rerpha: I've sorted it now by using `selectedItems()`

nexus_constructor/add_component_window.py
rerpha: Removing this disables selection completely. This doesn't have anything to do with `self.selected_widget`

nexus_constructor/component_fields.py
rerpha: I think that's a good idea, I'm just not sure the best way to go about doing it. Wouldn't you still be doing equality checks on the type of the class regardless of whether you use inheritance? I like the idea of separating them out as i think the field widget has too much responsibility currently.

rerpha:
I am going to make tickets for the suggestions in this ticket minus the polymorphism one, just had a go and couldn't get anywhere without weird behaviour.

nexus_constructor/component_fields.py
matthew-d-jones: I agree that might be the way to go when there are more `FieldType`s to deal with, but I think it will be easier to get the implementation of that right when that is the case rather than trying to do it now. I suggest the current implementaiton is less complex at this stage.

matthew-d-jones:
@rerpha I still see the combo box truncation problem that Dom pointed out. Could the minimum size of the name field be reduced slightly?

matthew-d-jones:
@rerpha  At some point in the history of this branch, when the name field got focus it would display the full list of suggested field names, can we get that behaviour back?

nexus_constructor/component_fields.py
DominicOram: Sure, sounds sensible.

PR #362, opened 2019-07-04
358: Add axes in the main view

### LINKED ISSUE #358: Add axes in the main 3D view
I think it would be good to display axes as thin lines in the main 3D view. They could be colour matched to the axes in the gnomon.

Example from CAD software CAD software

DolicaAkelloEgwel:
### Issue

Closes #358

### Description of work

- Added `InstrumentViewAxes` and `LineGeometry` classes
- Added a single test for `InstrumentViewAxes` (no tests for `LineGeometry`)
- Added a `AxisColors` enum so that it can be used by `Gnomon` and `InstrumentViewAxes`
- `set_material_properties` and `add_q_components_to_entity` moved to their own file because these functions are used by `InstrumentView`, `Gnomon`, and `InstrumentViewAxes`
- Also moved the tests, and added an extra test for `set_material_properties` (checking that shininess is set to zero in the case of the gnomon cylinders)
- ~~.flake8 now ignores E302 as this disagrees with black~~ (this doesn't seem to be needed now that qentity_utils has imports)
- Removed some method mocks from the tests that weren't actually needed
- Created separate entities for the axes lines and components so that the component entity is used with the camera's `viewAll' method and the combined entity is used to create the clear buffers
- Made a `InstrumentZooming3DWindow` class so that Esc only cares about the instrument components

Set the lines to be the length of the main camera's far plane thinking this would make it appear infinite but it didn't...

### Acceptance Criteria 

Check that the tests pass. Check that the axes lines look OK. Check the code because a few methods had to be moved/changed and I might have forgotten to update a docstring/comment somewhere.

### UI tests

n/a


DolicaAkelloEgwel:
Another thing is that pressing Esc now pretty much make everything disappear. In switch_to_widgets this would make the camera view all the components. Now that it wants to show the components plus the lines it has to zoom out a lot to the point it can't see anything.

nexus_constructor/instrument_view_axes.py
```diff
@@ -0,0 +1,93 @@
...
+class InstrumentViewAxes(object):
+    def __init__(self, component_root_entity: Qt3DCore.QEntity, line_length: float):
+        """
+        Class for housing the objects that create the axes in the instrument view.
+        :param component_root_entity: The root entity for the instrument view components.
+        :param line_length: The length of the line in the axes.
+        """
+
+        self.component_root_entity = component_root_entity
+        self.line_length = line_length
```
DominicOram: I don't think you need to hold either of these references.

nexus_constructor/instrument_view_axes.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # The coordinate array given to the LineGeometry must be in the form
+        # [start_x, start_y, start_z, end_x, end_y, end_z]. All lines start at the origin so the first three elements
+        # in the lists for the three lines will be the same.
+        origin = [0, 0, 0]
+        x_line_vertices = origin + [line_length, 0, 0]
+        y_line_vertices = origin + [0, line_length, 0]
+        z_line_vertices = origin + [0, 0, line_length]
```
DominicOram: This still seems to work fine without the origin part so I believe the comment is wrong

nexus_constructor/instrument_view_axes.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.x_line_mesh.setPrimitiveType(Qt3DRender.QGeometryRenderer.Lines)
+        self.y_line_mesh.setPrimitiveType(Qt3DRender.QGeometryRenderer.Lines)
+        self.z_line_mesh.setPrimitiveType(Qt3DRender.QGeometryRenderer.Lines)
+
+        self.x_line_mesh.setGeometry(self.x_line_geometry)
+        self.y_line_mesh.setGeometry(self.y_line_geometry)
+        self.z_line_mesh.setGeometry(self.z_line_geometry)
+
+        self.x_line_entity = Qt3DCore.QEntity(self.component_root_entity)
```
DominicOram: I don't think you need to hold these references.

nexus_constructor/instrument_view_axes.py
```diff
@@ -0,0 +1,93 @@
...
+
+        add_qcomponents_to_entity(
+            self.x_line_entity, [self.x_line_mesh, self.x_line_material]
+        )
+        add_qcomponents_to_entity(
+            self.y_line_entity, [self.y_line_mesh, self.y_line_material]
+        )
+        add_qcomponents_to_entity(
+            self.z_line_entity, [self.z_line_mesh, self.z_line_material]
+        )
```
DominicOram: In general this whole thing is just doing the same thing 3 different times, we should be able to get it into a loop. Maybe something like:
```
self.geometries = dict()
self.line_meshes = dict()
self.materials = dict()

for i, colour in enumerate([AxisColors.X.value, AxisColors.Y.value, AxisColors.Z.value]):
    # Calculates the end x, y and z for each line in turn
    vertices = [0] * 3
    vertices[i] = line_length

    self.geometries[i] = LineGeometry(
        QtCore.QByteArray(self.create_data_array(vertices))
    )
    self.line_meshes[i] = Qt3DRender.QGeometryRenderer()
    self.line_meshes[i].setPrimitiveType(Qt3DRender.QGeometryRenderer.Lines)
    self.line_meshes[i].setGeometry(self.geometries[i])

    self.materials[i] = Qt3DExtras.QPhongMaterial()

    set_material_properties(
        self.materials[i], colour, colour
    )

    add_qcomponents_to_entity(
        Qt3DCore.QEntity(component_root_entity), [self.line_meshes[i], self.materials[i]]
    )
```
But that sacrifices some readability, what do you think?

nexus_constructor/instrument_zooming_3d_window.py
```diff
@@ -0,0 +1,33 @@
...
+class InstrumentZooming3DWindow(Qt3DExtras.Qt3DWindow):
+    def __init__(self):
+        """
+        A custom 3D window that only zooms in on the instrument components when the escape key is pressed.
+        """
+        super().__init__()
+        self.component_root_entity = None
+        self.prev_position = None
+
+    def set_component_root_entity(self, component_root_entity):
```
DominicOram: You should pass this entity in as part of the constructor. Otherwise you risk calling `viewEntity(None)` below.

nexus_constructor/instrument_zooming_3d_window.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class InstrumentZooming3DWindow(Qt3DExtras.Qt3DWindow):
+    def __init__(self):
+        """
+        A custom 3D window that only zooms in on the instrument components when the escape key is pressed.
+        """
+        super().__init__()
+        self.component_root_entity = None
+        self.prev_position = None
```
DominicOram: `self.prev_position` isn't used anywhere?

nexus_constructor/line_geometry.py
```diff
@@ -0,0 +1,27 @@
...
+
+class LineGeometry(Qt3DRender.QGeometry):
+    def __init__(self, vertex_data: QtCore.QByteArray, parent=None):
+        """
+        Geometry for depicting a line. Used for axes placed in the origin of the instrument view.
+        :param vertex_data: A byte array containing the coordinates for the start and end points of the line.
+        :param parent: The object parent.
+        """
+        Qt3DRender.QGeometry.__init__(self, parent)
+        self.m_position_buffer = Qt3DRender.QBuffer(self)
```
DominicOram: Can you add some comments to describe what's going on here? Also, what does the `m_` prefix signify?

nexus_constructor/qentity_utils.py
```diff
@@ -0,0 +1,41 @@
+from typing import List
+
+from PySide2.Qt3DCore import Qt3DCore
+from PySide2.Qt3DRender import Qt3DRender
+from PySide2.QtGui import QColor
+
+
+def set_material_properties(
```
DominicOram: I think that in almost all the places you call this you basically do:

```
material = Qt3DExtras.QPhongMaterial()
set_material_properties(material)
```
Maybe it's worth moving the material creation into this method and streamline the code elsewhere.

nexus_constructor/qentity_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    material.setDiffuse(diffuse)
+
+    if alpha is not None:
+        material.setAlpha(alpha)
+
+    if remove_shininess:
+        material.setShininess(0)
+
+
+def add_qcomponents_to_entity(
```
DominicOram: As above you could potentially do similar here where you just pass the root entity and this deals with creating the new entity and then returns it.

PR #384, opened 2019-07-09
Refactor geometry models to wrap nexus file

### LINKED ISSUE #334: Widgets: add geometry information to component group in nexus file

matthew-d-jones:
### Issue

Closes #334

### Description of work

Geometry types for meshes, cylinders and placeholder shape have been reworked and now store their data in the in-memory NeXus file. The placeholder `NoShapeGeometry` does not get stored in file, as it should not end up in the output NeXus or JSON file.

The old `Component` and `Transformation` classes are now completely gone, so I've removed the `Model` suffix from the new `ComponentModel` and `TransformationModel` classes.

### Acceptance Criteria 

You should be able to add cylinders and mesh geometry from file in the add component interface and see them appear in the instrument view.


nexus_constructor/geometry/utils.py
```diff
@@ -0,0 +1,18 @@
+from PySide2.QtGui import QVector3D
+import numpy as np
+
+
+def validate_nonzero_qvector(value: QVector3D):
+    if value.x() == 0 and value.y() == 0 and value.z() == 0:
+        raise ValueError("Vector is zero length")
+
+
+def get_an_orthogonal_unit_vector(input_vector: QVector3D):
```
rerpha: Does this need a return type hint?

nexus_constructor/geometry/off_geometry.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        if problems:
+            raise NexusFormatError("\n".join(problems))
+
+    @property
+    def winding_order(self):
+        return [point for face in self.faces for point in face]
+
+    @property
+    def winding_order_indices(self):
```
rerpha: return type hint?

nexus_constructor/geometry/cylindrical_geometry.py
```diff
@@ -0,0 +1,208 @@
...
+            "NXcylindrical_geometry",
+            (
+                ValidateDataset(
+                    "vertices", shape=(None, 3), attributes={"units": None}
+                ),
+                ValidateDataset("cylinders", (3,)),
+            ),
+        )
+        if problems:
+            raise NexusFormatError("\n".join(problems))
```
rerpha: Could use a format string here? i'm not sure what's more pythonic.

nexus_constructor/component.py
```diff
@@ -245,16 +247,48 @@ def depends_on(self, transformation: TransformationModel):
...
+        Overrides any existing shape
+        """
+        self._remove_shape()
+        shape_group = self.file.create_nx_group("shape", "NXoff_geometry", self.group)
+        record_faces_in_file(self.file, shape_group, loaded_geometry.faces)
+        record_vertices_in_file(self.file, shape_group, loaded_geometry.vertices)
+        return OFFGeometryNexus(self.file, shape_group)
+
+    def get_shape(self) -> Optional[Union[OFFGeometry, CylindricalGeometry]]:
+        if "shape" in self.group:
```
rerpha: there are a few mentions of "shape" here. perhaps we should put this into a variable to avoid potential typos?

nexus_constructor/geometry/off_geometry.py
rerpha: if so, there are a couple of other properties in here which need it too.

nexus_constructor/geometry/off_geometry.py
```diff
@@ -0,0 +1,233 @@
...
+            for vertex_number in range(number_of_vertices)
+        ]
+
+    @vertices.setter
+    def vertices(self, new_vertices: List[QVector3D]):
+        record_vertices_in_file(self.file, self.group, new_vertices)
+
+    @property
+    def faces(self) -> List[List[int]]:
+        winding_order_from_file = self.group["winding_order"]
```
DominicOram: It's not immediately obvious what this function is doing. Can we have some tests that specifically cover this method?

nexus_constructor/geometry/off_geometry.py
```diff
@@ -0,0 +1,233 @@
...
+        ]
+        faces.append(winding_order_from_file[face_starting_indices[-1] :].tolist())
+        return faces
+
+    @faces.setter
+    def faces(self, new_faces: List[List[int]]):
+        record_faces_in_file(self.file, self.group, new_faces)
+
+
+def record_faces_in_file(
```
DominicOram: Can this be covered by some unit tests?

nexus_constructor/nexus/validation.py
```diff
@@ -0,0 +1,90 @@
...
+import h5py
+from typing import Tuple, List, Optional
+
+
+class NexusFormatError(Exception):
+    pass
+
+
+@attr.s
+class ValidateDataset:
```
DominicOram: Can we have some unit tests to cover this?

tests/test_component.py
```diff
@@ -1,29 +1,21 @@
...
+)
 
 
 def test_can_create_and_read_from_field_in_component():
     field_name = "some_field"
     field_value = 42
-    nexus_wrapper = NexusWrapper(str(uuid1()))
-    component_group = _add_component_to_file(nexus_wrapper, field_name, field_value)
-    component = ComponentModel(nexus_wrapper, component_group)
+    nexus_wrapper = create_nexus_wrapper()
```
DominicOram: If we're creating this wrapper and component for every test does it make sense just to do it in a fixture?

tests/test_cylindrical_geometry.py
```diff
@@ -0,0 +1,28 @@
+from pytest import approx, raises
+from PySide2.QtGui import QVector3D
+from .helpers import create_nexus_wrapper, add_component_to_file
+
```
DominicOram: I think in general there should be more tests in here, for example:
* Changing properties of the cylinder
* Loading a cylinder from a file where it has multiple cylinders or cylinders that are not `[0,1,2]`
* That the OFF geometry  of the cylinder is correctly computed
* That the underlying nexus file changes correctly
* Probably many others, worth having a think

nexus_constructor/geometry/off_geometry.py
```diff
@@ -0,0 +1,233 @@
...
+    @property
+    def faces(self) -> List[List[int]]:
+        return self._faces
+
+    @faces.setter
+    def faces(self, new_faces: List[List[int]]):
+        self._faces = new_faces
+
+
+class OFFGeometryNexus(OFFGeometry):
```
DominicOram: Can you add some unit tests to cover this class in general?

matthew-d-jones: It is tested in `test_off_geometry.py`. Maybe it isn't obvious because I populate it by using an `OFFGeometryNoNexus`?

tests/test_component.py
matthew-d-jones: I did consider it, but in this case I feel the setup is relevant to the tests rather than just being required boilerplate, so I'm not too keen on "hiding it" in a fixture. I opted instead for the helper functions, which reduce it to a couple of lines whilst still making to clear what the scenario for the test is.

nexus_constructor/geometry/cylindrical_geometry.py
```diff
@@ -0,0 +1,208 @@
...
+)
+from nexus_constructor.geometry.utils import (
+    validate_nonzero_qvector,
+    get_an_orthogonal_unit_vector,
+)
+from nexus_constructor.geometry.off_geometry import OFFGeometry, OFFGeometryNoNexus
+from typing import Tuple
+
+
+def calculate_vertices(
```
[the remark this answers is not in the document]
matthew-d-jones: It's covered, but not directly, agree it is worth direct tests as documentation.

nexus_constructor/geometry/off_geometry.py
matthew-d-jones: It's covered, but not directly, agree it is worth direct tests as documentation.

nexus_constructor/nexus/validation.py
matthew-d-jones: It's covered, but not directly, agree it is worth direct tests as documentation.

nexus_constructor/geometry/cylindrical_geometry.py
```diff
@@ -0,0 +1,208 @@
...
+        return cylinder_axis.normalized()
+
+    @axis_direction.setter
+    def axis_direction(self, new_axis_direction: QVector3D):
+        """
+        Not required to be a unit vector, but must be non-zero
+        :param new_axis_direction: Vector describing direction of the cylinder's axis
+        """
+        validate_nonzero_qvector(new_axis_direction)
+        # TODO rotate the vertices by the angle between the new and old axis directions
```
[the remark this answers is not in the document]
matthew-d-jones: It seems a setter for that field it isn't needed currently, so I'll remove the method entirely. I'm too used to using python for throw-away scripts rather than application dev; need to get into the habit of TDD for this too.

nexus_constructor/geometry/cylindrical_geometry.py
```diff
@@ -0,0 +1,208 @@
...
+        self.group["vertices"].attrs["units"] = new_units
+
+    @property
+    def height(self) -> float:
+        base_centre, _, top_centre = self._get_cylinder_vertices()
+        cylinder_axis = top_centre - base_centre
+        return cylinder_axis.length()
+
+    @height.setter
+    def height(self, new_height: float):
```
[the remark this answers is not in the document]
matthew-d-jones: Unused, have removed the method

nexus_constructor/geometry/off_geometry.py
matthew-d-jones: Added `test_can_retrieve_list_of_vertices_for_each_face` to `test_off_geometry.py`

nexus_constructor/geometry/off_geometry.py
matthew-d-jones: Added several more tests to `test_off_geometry`

nexus_constructor/geometry/off_geometry.py
matthew-d-jones: added test: `test_can_record_list_of_vertices_for_each_face` in `test_off_geometry.py`

nexus_constructor/nexus/validation.py
matthew-d-jones: added tests in `test_nexus_validation.py`

tests/test_cylindrical_geometry.py
matthew-d-jones: The setters were not being used, so I've removed them. When we can edit components a cylinder may need to change, but they are as cheap to remove and recreate as to change, so I think keeping them immutable is more simple.

I've added various tests, including for out-of-order vertices and to document what happens if there are multiple cylinders defined in the group in file.

I've not tested the OFF geometry output for cylinders as this is not a regression for these changes and that functionality will soon be replaced: #397 (I don't think it is worth my time.)

matthew-d-jones:
@rerpha @DominicOram 
I marked your comments as resolved as I worked through them, feel free to "unresolve" them if you think more work is needed.

tests/test_component.py
DominicOram: Sure, that's reasonable.

rerpha:
@DominicOram we should have a think about whether we use that for everything or not in the next NC planning meeting and if so make a ticket to rename the tests to fit that format

PR #392, opened 2019-07-10
Allow spinboxes in cylinder editing section to take doubles/floats

### LINKED ISSUE #387: Cylinder spinboxes only accept integers
A user is not able to enter values with decimal points due to the spinboxes in the add component window.

rerpha:
### Issue

Closes #387 

### Description of work

- Replaced `QSpinBox` instances in `add_component.ui` with `QDoubleSpinBox`
- Re-generated the python file from the .ui file

### Acceptance Criteria 

Cylinders now take floats as well as integers. 

### UI tests

N/A

### DIFF
--- a/ui/add_component.py
+++ b/ui/add_component.py
@@ -3,7 +3,7 @@
 # Form implementation generated from reading ui file 'add_component.ui',
 # licensing of 'add_component.ui' applies.
 #
-# Created: Thu Jun 13 16:16:47 2019
+# Created: Wed Jul 10 10:28:16 2019
 #      by: pyside2-uic  running on PySide2 5.12.3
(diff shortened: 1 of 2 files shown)


rerpha:
> Do you think anyone might want more than two decimal places?

That's a good point, i hadn't thought about that. I think it's easy enough to change if needs be

rerpha:
thinking about it, these are all distances - I think 2 will do for now, if a user wants more precision they could step down a unit (m -> cm etc)

matthew-d-jones:
Precision for floating point numbers is a general problem that we need to decide how to handle across the application. I think 2 decimal places is fine here for now.

Issue #313, opened 2019-06-10
Add GUI tests


matthew-d-jones:
After switch to widgets we should now be able to use pytest-qt to automate testing of the GUI.

rerpha:
Try rolling back to 5.12.1 - UI tests were working then

rerpha:
Closed in #400

PR #400, opened 2019-07-12
313: Add GUI tests

DolicaAkelloEgwel:
### Issue

Makes progress on #313.

### Description of work

Adds some GUI tests for the `AddComponentDialog`. These mainly involve checking that the right boxes appear/disappear, that fields become red/white when they ought to, that the Add Component button is enabled/disabled. Also added some tests that check the `AddComponentDialog` closes when a user enters valid input and presses the Add Component button. 

No tests related to the fields items have been introduced yet. Also no tests that involve checking that a new item is added to the component list after adding a component. Have found a way to make `QRadioButton` presses work on Windows and Linux. 

In the case of the first test I found out that you can call `component_tool_bar.widgetForAction` and use this as an argument for `mouseClick` in order to interact with `QAction`s as though they were widgets. Using this with the add component `QAction` makes `AddComponentDialog` appear without the need for a position argument but the test still gets a segfault so it will have to remain skipped for now.

### Acceptance Criteria 

Check that the tests pass and that appropriate test cases have been covered.

### UI tests

n/a


rerpha:
have fixed some of the tests, have left the old code as comments, up to you whether you want to remove. I think maybe we should have a look around and see if anyone else has had a similar experience with pytest-qt on windows being slightly different. I wonder if the event handlers work differently depending on your window manager.

rerpha:
I think we can remove the `ui/` directory ignore in the codecov.yml config, as those should be covered with some tests?

PR #401, opened 2019-07-15
Fields arrays

### LINKED ISSUE #331: Add field dialog for Array datasets in fields widget
Add a dialog that allows a user to enter array values (a table of some sort) to fields in the component group. This should be opened when the "edit" button is pressed on an item in the fields list widget.

rerpha:
### Issue

Closes #331 

### Description of work

Adds a dialog and functionality for adding array fields to a nexus file. 

### Acceptance Criteria 

- test with each dtype that the array ends up in the nexus file in the correct place

### UI tests

Added some UI tests for the tableview that has been added.


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E203, E266, E501, W503, E123, C901
+ignore = E203, E266, E501, W503, E123, C901, W504
```
matthew-d-jones: Can we use a `# noqa: W504` comment on the troublesome lines rather than add it to the ignore list?

tests/test_validators.py
```diff
@@ -259,3 +261,27 @@ def test_GIVEN_invalid_file_WHEN_using_ok_validator_WITH_mesh_button_unchecked_T
...
+    assert validator.validate("1", 0) == QValidator.Acceptable
+    validator.is_valid.emit.assert_called_once_with(True)
+
+
+def test_GIVEN_floating_point_value_WHEN_using_numpy_validator_with_integer_as_dtype_THEN_false_signal_is_emitted():
+    validator = NumpyDTypeValidator(np.intc)
+    validator.is_valid = Mock()
+
+    assert validator.validate("1.2", 0) == QValidator.Intermediate
+    validator.is_valid.emit.assert_called_once_with(False)
```
matthew-d-jones: Could add a test for the validator for alphabetical chars with float type?

matthew-d-jones:
@rerpha Elements are changing position when I add columns, I would guess the reshape needs to be done in 'F' mode when adding columns (and left as the default 'C' mode when adding rows).

matthew-d-jones:
@rerpha Remove row and column don't appear to function if the currently selected cell is not in the last row/column of the table.

rerpha:
the remove row/column behaviour should now disable the user from being able to delete a column or row if it is the last one left but should not stop deleting the other axes element if there are still more than one elements

Issue #404, opened 2019-07-16
Add UI tests for main window


DolicaAkelloEgwel:

rerpha:
We should have a think about what we want to test for the main window, a couple of ideas from me are: 

- file dialogs opening when the actions/buttons in the `file` menu are selected
- pressing the nexus file layout tab button displays the silx treeview
- buttons that should be enabled in the toolbar for the component view
- expanding an item in the components

i can't really think of a good way to test the 3D view, maybe if there's an easy way to see if the sample shape is loaded, but i think this is already covered in other unit tests?

rerpha:
I don't think we're able to do this for now as the 3d window even if overridden causes pytest to segfault. Marking as blocked for now.

DominicOram:
See if this is fixed after https://github.com/ess-dmsc/nexus-constructor/issues/377

rerpha:
Still an issue, even in pyside2 5.13.1.

DominicOram:
`main_window.py` is growing and some of the logic in there is quite horrible (see `set_button_state`). If this ticket is blocked for the foreseeable future can we change this ticket to instead refactor `main_window.py` to separate out the parts that are testable and then just test them?

kmurica:
@matthew-d-jones I am curious if we can't just close this ticket? It seems obsolete. If tests are needed to be added after some additional refactoring and/or GUI development, we should create new tickets in Jira or issues here.

matthew-d-jones:
Agreed

PR #407, opened 2019-07-17
Replacing line edits with spin boxes for transformation frames

### LINKED ISSUE #376: Implement proper verification of inputs of the transformation edit frames of the component view
As the title says.

rerpha:
### Issue

Closes #376 

### Description of work

Just thought i'd do this really quickly as it'll make the program quite a bit more robust.

I have replaced the line edits in the transformation frames so that a user cannot input anything other than numerical values, previously they could put "asdfisdfi" and it would still work (or do nothing) 

no more casting from float to str! and vice versa. 

### Acceptance Criteria 

Jenkins passes
Check no errors occur when adding transformations 

### UI tests

I haven't added any, because the UI for this part of the program is likely about to change anyway. #404 should cover the transformations stuff when done. 

### DIFF
--- a/ui/transformation.py
+++ b/ui/transformation.py
@@ -3,8 +3,8 @@
 # Form implementation generated from reading ui file 'transformation.ui',
 # licensing of 'transformation.ui' applies.
 #
-# Created: Thu Jul  4 23:55:45 2019
-#      by: pyside2-uic  running on PySide2 5.12.3
+# Created: Wed Jul 17 14:07:33 2019
+#      by: pyside2-uic  running on PySide2 5.13.0
 #
 # WARNING! All changes made in this file will be lost!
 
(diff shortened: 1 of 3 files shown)


ui/transformation.py
```diff
@@ -54,19 +54,19 @@ def setupUi(self, Transformation):
         self.xPosLabel = QtWidgets.QLabel(Transformation)
         self.xPosLabel.setObjectName("xPosLabel")
         self.vectorLayout.addWidget(self.xPosLabel)
-        self.xLineEdit = QtWidgets.QLineEdit(Transformation)
+        self.xLineEdit = QtWidgets.QDoubleSpinBox(Transformation)
```
DolicaAkelloEgwel: I think these should name their names changed now.

PR #421, opened 2019-07-29
306: Add pixel controls

### LINKED ISSUE #306: Widgets: add pixel controls
depends on #303 

Add the pixel controls in the dialog which add the models for pixel states as well as the listviews in the add component dialog.

DolicaAkelloEgwel:
### Issue

Closes #306 

### Description of work

- The pixel-related UI elements are contained in their own widget. The widget is is managed by `PixelOptions`.
- The AddComponentDialog interacts with the `PixelOptions` and tells it when to populate the Pixel Mapping list, retrieves information from it when Add Component is pressed, etc.
- The pixel options automatically appear if the the shape is Cylinder/Mesh and the component type is `NXdetector` or `NXdetector_module` selected allow pixel-related input in the NeXus file.
- The pixel options widget has three radio buttons that when clicked control what input fields appear: Single Pixel, Entire Shape, No Pixels.
- "Single Pixel" is selected by default.
- The "Single Pixel" button allows users to enter a Pixel Grid.
    - If the row/column count is set to zero, then the matching row height/column width field is disabled.
    - If the user presses Add Component when either the row height or column width is disabled, then its value in the interface before the spin box was disabled is ignored. The `PixelGrid` object is created with a value of zero for the row height/column width.
    - If both the row count or a column count are zero then the Add Component button is disabled and the spin boxes for row/column count turn red.
    - There is no way of entering units for the row height/column width. This is in #425.
- The "Single Pixel" button allows users to enter a Pixel Grid.
    - A valid pixel mapping has at least one ID given. If none are given then the Add Component button is disabled.
    - In the case of `NXcylindrial_geometry` the Pixel Mapping can only record a single ID. Allowing a user to enter multiple cylinders is now in #429.
- Nothing happens for the pixel data given to a `NXdetector_module` (but the info does still get sent to the `NXoff_geometry` or `NXcylindrical_geometry`). This is now in #424.
- The pixel data can't be edited. This is in #430.

Also I'm wondering if it might be worth switching over to `QWizard` if the Add Component stuff becomes more complicated? 

### Acceptance Criteria 

- Check that the tests pass and that appropriate test cases are covered
- Try entering some pixel info and then saving to a NeXus file and checking that what you see in HDFView matches what you gave to the NeXus Constructor

### UI tests

n/a


tests/ui_tests/test_ui_add_component_window.py
```diff
@@ -827,6 +1134,349 @@ def test_UI_GIVEN_array_field_selected_and_edit_button_pressed_THEN_edit_dialog_
     assert field.table_view.isEnabled()
 
 
+def test_UI_GIVEN_user_selects_pixel_grid_THEN_pixel_grid_is_set_to_true_in_ok_validator(
```
DominicOram: There's a lot of repeated code in the following 10 tests. Don't be afraid to create functions to do the common bits

tests/ui_tests/test_ui_add_component_window.py
```diff
@@ -1000,3 +1649,84 @@ def test_UI_GIVEN_field_widget_with_string_type_THEN_value_property_is_correct(q
...
+    :param dialog: An instance of an AddComponentDialog.
+    :param template: The window/widget that holds the AddComponentDialog.
+    :param component_name: The desired component name.
+    """
+    qtbot.mouseClick(dialog.nameLineEdit, Qt.LeftButton)
+    qtbot.keyClicks(dialog.nameLineEdit, component_name)
+    show_and_close_window(qtbot, template)
+
+
+def systematic_button_press(
```
DominicOram: These are generic helper functions for ui tests, they should be in a separate file.

DominicOram:
I've had a quick look through this, it looks good. However, there's a lot of changes, the ticket should probably have been split up a bit. In general:
* Can you split the pixel stuff out off `add_component_window.py`? As it stands that file is growing quickly and has too much responsibility. 
* Also split the pixel validator stuff out of the OK validator
* The number of new tests is great, but again we're getting some rather large files now, can we split them into files that test different aspects?

nexus_constructor/component_type.py
```diff
@@ -37,7 +37,7 @@
     "NXslit",
     "NXvelocity_selector",
     "NXxraylens",
-]
```
DolicaAkelloEgwel: Explanation for this change: A test I wrote earlier was failing because `NXdetector_module` was placed in this list twice. This caused a mismatch between the length of the list and the number of elements in the drop-down list in the Add Component Dialog. Changed the list to a set and removed the extra `NXdetector_module` to be safe.

matthew-d-jones:
> Also I'm wondering if it might be worth switching over to QWizard if the Add Component stuff becomes more complicated?

Yes, I was also thinking along those lines, there may even be a related ticket. The real challenge will be working out how to map pixel IDs to the more exotic detector geometries.

nexus_constructor/validators.py
```diff
@@ -153,18 +172,63 @@ def open_file(filename: str, mode: str = "r"):
     is_valid = Signal(bool)
 
 
+class PixelValidator(QObject):
```
matthew-d-jones: A doc comment to describe what this validator does would be good to have.

tests/test_nexus_wrapper.py
```diff
@@ -1,12 +1,36 @@
...
 
+global file_count
+file_count = 0
+
+
+@pytest.fixture(scope="function")
+def nexus_file():
+    global file_count
+    file_count += 1
+    return create_in_memory_file("test_nw" + str(file_count - 1))
```
matthew-d-jones: Rather than accumulating many in-memory files and having to give them unique names you could just close the file at the end of each test. It can be achieved something like this (I didn't test the code below):
```python
@pytest.fixture(scope="function")
def nexus_file(request):
    test_file = create_in_memory_file("test_nw")
    def fin():
        test_file.close()
    request.addfinalizer(fin)
    return test_file
```

tests/test_pixel_data_to_nexus_utils.py
```diff
@@ -0,0 +1,138 @@
...
+def test_GIVEN_list_of_ids_WHEN_calling_detector_number_THEN_correct_detector_number_list_is_returned(
+    pixel_mapping
+):
+
+    expected_numbers = [id for id in pixel_mapping.pixel_ids if id is not None]
+
+    assert detector_number(pixel_mapping) == expected_numbers
+
+
+@pytest.mark.parametrize("rows", ROW_COL_VALS)
```
matthew-d-jones: :+1:

nexus_constructor/pixel_data_to_nexus_utils.py
```diff
@@ -0,0 +1,78 @@
...
+def pixel_grid_detector_ids(grid: PixelGrid):
+    """
+    Returns a list of 'row' lists of 'column' length.
+    Each entry in the sublists are detector id's of pixel instances in the given PixelGrid
+    """
+
+    ids = np.arange(grid.rows * grid.columns)
+
+    if grid.count_direction == CountDirection.COLUMN:
+        ids = ids.reshape((grid.rows, grid.columns), order="F")
```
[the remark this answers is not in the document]
matthew-d-jones: >  ‘F’ means to read / write the elements using Fortran-like index order, with the first index changing fastest, and the last index changing slowest.

nexus_constructor/add_component_window.py
```diff
@@ -284,31 +333,107 @@ def generate_geometry_model(self, component: Component) -> OFFGeometry:
...
     def on_ok(self):
         nx_class = self.componentTypeComboBox.currentText()
         component_name = self.nameLineEdit.text()
         description = self.descriptionPlainTextEdit.text()
 
+        pixel_data = self.pixel_options.generate_pixel_data()
+        pixel_data_is_mapping = type(PixelData) is PixelMapping
+
         if self.component_to_edit:
             self.component_to_edit.name = component_name
```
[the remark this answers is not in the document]
matthew-d-jones: @DolicaAkelloEgwel this is still todo

PR #426, opened 2019-08-06
using qcylindermesh instead of OFFmesh for cylinders

### LINKED ISSUE #397: Render cylinders as cylinders in the 3D view
Instead of generating a mesh for cylinders, use the higher level interface of qt3d and simply tell it to render a cylinder.

The `off_geometry` method can be removed from `OFFGeometry` and `CylindricalGeometry`.

rerpha:
### Issue

Closes #397 
### Description of work

Only spent a few mins on this so if it causes issues i'll come back to it at a later date.
This uses `QCylinderMesh`es instead of drawing cylinders as OFF meshes in the qt3d view. This may give some performance benefits, but mainly it reduces a lot of code that we have to look after (converting cylinders to off meshes) 

### Acceptance Criteria 

Cylinders should be added as usual in the 3d view and the same output should be shown in the nexus file and tree view. 

### UI tests

*How you have updated the UI tests document to fit your changes*

### DIFF
--- a/nexus_constructor/geometry/off_geometry.py
+++ b/nexus_constructor/geometry/off_geometry.py
@@ -9,6 +9,7 @@ from nexus_constructor.nexus.validation import (
     ValidateDataset,
     validate_group,
 )
+from nexus_constructor.off_renderer import OffMesh
 from nexus_constructor.ui_utils import (
     numpy_array_to_qvector3d,
     qvector3d_to_numpy_array,
@@ -38,7 +39,7 @@ class OFFGeometry(ABC):
 
     @property
     @abstractmethod
-    def off_geometry(self) -> "OFFGeometry":
+    def mesh(self) -> OffMesh:
         pass
 
     @property
@@ -89,8 +90,8 @@ class OFFGeometryNoNexus(OFFGeometry):
         return [sum(face_sizes[0:i]) for i in range(len(face_sizes))]
 
     @property
-    def off_geometry(self) -> OFFGeometry:
-        return self
+    def mesh(self) -> OffMesh:
+        return OffMesh(self)
 
     @property
     def vertices(self) -> List[QVector3D]:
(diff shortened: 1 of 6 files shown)


nexus_constructor/geometry/off_geometry.py
```diff
@@ -38,7 +39,7 @@ def winding_order_indices(self) -> List[int]:
 
     @property
     @abstractmethod
-    def off_geometry(self) -> "OFFGeometry":
+    def mesh(self) -> OffMesh:
```
matthew-d-jones: I think the return type should be `Union(Qt3DExtras.QCylinderMesh, OFFMesh)`

matthew-d-jones: Or if `OFFMesh` and `QCylinderMesh` inherit from a base mesh class then that would be appropriate

nexus_constructor/off_renderer.py
```diff
@@ -7,11 +7,14 @@
...
 from nexus_constructor.pixel_data import PixelData, PixelGrid
-from nexus_constructor.geometry import OFFGeometry
 from PySide2.Qt3DRender import Qt3DRender
 from PySide2.QtGui import QVector3D
 import struct
 import itertools
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover
+    from nexus_constructor.geometry.off_geometry import OFFGeometry
```
matthew-d-jones: As you've changed to strings for the OFFGeometry type hint, surely this import doesn't do anything?

rerpha: this is to suppress flake8 warnings, but shouldn't need to exist if i make them inherit from a base class

rerpha:
I am a bit stuck with this, have tried a few different methods such as bundling in a transform with the cylinder, but i'm not sure how to get it finished. I am going to shelve for now.

rerpha:
Closing for now as the branch will diverge with upcoming changes and I can't find a straightforward way of fixing this.

PR #432, opened 2019-08-12
377: Fix crash on exit (segfault)

### LINKED ISSUE #377: Fix crash on exit
The application crashes on exit on my machine. It seems likely that this is caused by some python resource is being freed before the C++ Qt backend is done with it. Though currently only a nuisance, it should be fixed.

rerpha:
### Issue

Closes #377 

### Description of work

Fixes the segfault caused by python not clearing stuff up correctly with qt3d. 

Deleted the entities before qt3d tries to reference them to draw the next frame in between closing the application. 

### Acceptance Criteria

- The program exits with  exit code 0 instead of `SIGSEGV`

### UI tests

N/A

### DIFF
--- a/nexus_constructor/instrument_view.py
+++ b/nexus_constructor/instrument_view.py
@@ -24,7 +24,9 @@ class InstrumentView(QWidget):
     """
 
     def delete(self):
-        """Fixes Qt3D segfault"""
+        """
+        Fixes Qt3D segfault - this needs to be called when the program closes otherwise Qt tries to draw objects as python is cleaning them up.
+        """
         self.clear_all_components()
         del self.root_entity
         del self.view
--- a/nexus_constructor/instrument_view.py
+++ b/nexus_constructor/instrument_view.py
@@ -4,7 +4,7 @@ from PySide2.Qt3DCore import Qt3DCore
 from PySide2.Qt3DExtras import Qt3DExtras
 from PySide2.Qt3DRender import Qt3DRender
 from PySide2.QtCore import QRectF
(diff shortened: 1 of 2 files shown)


nexus_constructor/instrument_view.py
```diff
@@ -23,6 +23,12 @@ class InstrumentView(QWidget):
                    argument in order to appease Qt Designer.
     """
 
+    def __del__(self):
```
SkyToGround: It is my understanding that Python does not give any guarantees about `__del__` being called. (Do you have any wise words here @matthew-d-jones ?)

On the other hand, if it works, why not use the solution that we have?

SkyToGround:
I did some testing on my machine and was able to determine that `__del__` is indeed not reliable called on exit. When it is called, the application exits with exit code 0. When `__del__` is not called, the application crashes.

nexus_constructor/instrument_view.py
matthew-d-jones: Yes `__del__` is not the right place to do this, it only gets called when/if the garbage collector deals with the object.

matthew-d-jones: Calling this code in the application exit function would be better.

rerpha:
oops didn't mean to close this. You can try and override closeEvent (from QWidget) but this does nothing, never gets called, and still crashes. I think the problem is that python is clearing the object up quicker than C++, therefore connecting to a C++ signal won't do anything and will not get called because it'll segfault first due to C++ referencing python objects that no longer exist.

PR #437, opened 2019-08-19
388: Draw disc chopper in Instrument View

### LINKED ISSUE #388: Draw chopper disc in instrument view
Use @DolicaAkelloEgwel's code in https://github.com/nexusformat/features/pull/34 to draw chopper discs based on their NeXus description (without requiring an OFF description). If there is a OFF or cylinder-based shape description present then it should take precedence.

matthew-d-jones:
### Issue

Closes #388 

### Description of work

Adds the disc chopper feature to the NeXus constructor. 

### Acceptance Criteria 

Check that the tests pass. Load a NeXus file that contains a disk chopper and create a chopper manually using the fields widgets. Check that in both cases the expected disc chopper appears.

### UI tests

n/a


nexus_constructor/geometry/disk_chopper_geometry.py
```diff
@@ -0,0 +1,497 @@
...
+CONTAINS_LONG = [DATASET_TYPE[key] for key in DATASET_TYPE.keys() if "Long" in key]
+INT_TYPES = CONTAINS_INTEGER + CONTAINS_SHORT + CONTAINS_LONG
+
+FLOAT_TYPES = [
+    DATASET_TYPE[key] for key in DATASET_TYPE.keys() if key in ["Float", "Double"]
+]
+
+
+class ChopperChecker:
+    def __init__(self, fields_widget: QListWidget):
```
matthew-d-jones: I think some refactoring will be needed here. Probably `ChopperChecker` shouldn't know about the field widget because this validation also needs to be carried out when loading a NeXus file, not just when using the Add Component window.

nexus_constructor/geometry/disk_chopper_geometry.py
```diff
@@ -0,0 +1,497 @@
...
+                return False
+
+        return True
+
+    def input_describes_valid_chopper(self):
+
+        self._slit_edges = self.fields_dict[SLIT_EDGES].value
+        self._radius = self.fields_dict[RADIUS].value
+        self._slit_height = self.fields_dict[SLIT_HEIGHT].value
+        self._slits = self.fields_dict[SLITS].value
```
matthew-d-jones: Could you construct a `ChopperDetails` at this point instead of duplicating these data as both members of `ChopperChecker ` and `ChopperDetails`?

matthew-d-jones:
I tried adding a chopper in the "Add Component" window but only see a black cube in the instrument view:
Screenshot_20190819_105749

rerpha:
Tabulate is fixed in the latest version (0.8.5) so you can revert the version pinning in `requirements.txt`

nexus_constructor/geometry/disk_chopper/disk_chopper_geometry_creator.py
```diff
@@ -0,0 +1,394 @@
+from typing import List, Tuple
+
+import numpy as np
+from PySide2.QtGui import QVector3D
+
+from nexus_constructor.geometry import OFFGeometryNoNexus
+from nexus_constructor.geometry.disk_chopper.chopper_details import ChopperDetails
+
+ARROW_SIZE = 0.1
```
matthew-d-jones: It would be good to have a comment for each of these constants to explain what they are used for.

nexus_constructor/geometry/disk_chopper/disk_chopper_geometry_creator.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    Tool for creating OFF Geometry in the form of strings from NXdisk_chopper information.
+    """
+
+    def __init__(self, chopper_details: ChopperDetails):
+
+        self.points = []
+        self.faces = []
+        self.arrow_size = ARROW_SIZE
+        self.resolution = RESOLUTION
```
matthew-d-jones: Do `ARROW_SIZE` and `RESOLUTION` really need to be copied as members?

Issue #445, opened 2019-09-18
Cylinder height and radius cannot be smaller than 1


matthew-d-jones:
If I select _Cylinder_ shape in the _Add Component_ it does not let me enter a height or radius smaller than `1.0`.

rerpha:
this is to do with the minimum values in the ui files. will make a PR as this is a 30 second fix

PR #448, opened 2019-09-30
Fixing floating point array field UI issues 

### LINKED ISSUE #412: Improve UI when editing floats in array fields
The behaviour when editing floats in an array field table is awkward, for example try typing `1.2` and you get `1.0.2`.

rerpha:
### Issue

Closes #412 

### Description of work

Quick and easy fix - fixes the weird behaviour when typing doubles in the array fields editor. This was due to the validator validating every time a character was entered, and rounding up to the nearest integer. To solve this, I have connected the `editingFinished` signal of the lineEdit to the validate function. This works even if the window is closed as focus is lost. 

### Acceptance Criteria 

Test entering floating point values as a `double` type in the array fields editing box. Previously this would replace your input of `1.2` with `1.0.2`. 

Also test closing the edit box when text has been entered into a cell - when opening the edit window it should still be there. 

### UI tests

N/A

### DIFF
--- a/nexus_constructor/array_dataset_table_widget.py
+++ b/nexus_constructor/array_dataset_table_widget.py
@@ -157,7 +157,7 @@ class ValueDelegate(QItemDelegate):
         super().__init__(parent)
         self.dtype = dtype
 
-    def commit(self, editor, _):
+    def commit(self, editor):
         """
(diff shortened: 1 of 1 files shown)


matthew-d-jones:
We should check if this also solves #445

PR #450, opened 2019-09-30
Making the 3d view resizable

### LINKED ISSUE #444: Allow resizing 3D view and tabs in main interface
Whilst using the application I've found it awkward that the divider between 3D view and tab view in the main interface is not movable/resizable. For example it would be useful to be able to see all columns in the file layout tab without horizontal scrolling.

rerpha:
### Issue

Closes #444 

### Description of work

Adds a qsplitter within the central grid layout so the 3d view can be resized on a splitter with the tab view. 

I have been able to do this through the generated python files but not the .ui files. I wonder if we should actually keep using these, as they don't seem that beneficial besides being compatible with qt designer. I have only used them once or twice since designing the layout initially, so i vote to get rid of them, as they are high maintenance and not adding any value besides a WYSIWYG editor which doesn't really work with the custom widgets anyway. 

### Acceptance Criteria 

Add qsplitter. We can configure to make the children collapsible or not, depending on preference. 

### UI tests

n/A


matthew-d-jones:
Let's discuss this at the next sprint meeting. I'm not against dropping the .ui file but if we do that we need to refactor the .py file to make it easier to read.

matthew-d-jones:
Merging as passed on CentOS, Mac and Windows nodes are down.

PR #462, opened 2019-10-07
361 write forwarder commands 

### LINKED ISSUE #361: Output forwarder command message
We should add a way to output a forwarder command message by looking through Kafka stream fields and selecting all components that are marked with `f142` as the schema.

### LINKED ISSUE #461: Add channel_provider_type and multiple converters to f142 stream options
only pva(v4) PVs will currently work in the JSON command because they are the default provider type in the forwarder - there is no way to add a `channel_provider_type` field for each PV or a way to add multiple converters for sending PVs to multiple topics etc.

rerpha:
### Issue

Closes #361 
Closes #461 

### Description of work

Adds a way to output an f142 add command from the streams that exist in the nexus constructor. 

I have created #461 to follow on from this. 

### Acceptance Criteria 

Added UI for another action in `main_window.py` and added a few functions in `writer.py` to dump the f142 data into a command. 

### UI tests

No UI tests as this is just an action, however functional tests have been added. 


nexus_constructor/nexus_filewriter_json/writer.py
```diff
@@ -204,3 +206,56 @@ def create_writer_commands(
         stop_cmd["stop_time"] = stop_time
 
     return write_cmd, stop_cmd
+
+
+def generate_forwarder_command(
```
matthew-d-jones: I don't think there is any reason to put these functions in this file.
Let's rename `nexus_filewriter_json` directory to something more general, maybe just `json`, and then rename `writer.py` to `filewriter_json_writer.py` or similar.

tests/test_nexus_to_json.py
```diff
@@ -330,3 +332,124 @@ def test_GIVEN_string_list_WHEN_getting_data_and_type_THEN_returns_correct_dtype
...
+
+    assert len(streams_) == 1
+
+    assert isinstance(streams_[0]["converter"], list)
+    assert streams_[0]["channel"] == pv_name
+    assert streams_[0]["converter"][0]["topic"] == topic1
+    assert streams_[0]["converter"][1]["topic"] == topic2
+
+
+def test_GIVEN_stream_with_pv_forwarding_to_three_topics_WHEN_generating_forwarder_command_THEN_stream_is_added_to_converters():
```
matthew-d-jones: What does this test that the previous test doesn't?

nexus_constructor/nexus_filewriter_json/writer.py
```diff
@@ -204,3 +206,56 @@ def create_writer_commands(
...
+    file, streams: Dict[str, Dict[str, Any]], provider_type: str
```
matthew-d-jones: `file` is a keyword, don't use it for variable name! Maybe `output_file`? Also needs type hint.
Doc comment with `params` description would also be useful.

tests/test_nexus_to_json.py
rerpha: Covers this branch: 
https://github.com/ess-dmsc/nexus-constructor/pull/462/files#diff-3b148aa9d555ba73e4949e927cd13dd9R55-R56

PR #463, opened 2019-10-07
Display pixel grids in 3D view

### LINKED ISSUE #434: Display pixel grids in 3D view
For an NXdetector with a pixel grid defined, the 3D view should show each pixel in the grid. Currently it only displays the single pixel shape.

The 3D view will need to display the same pixel shape at different positions by constructing a translation for each pixel from the pixel offsets.

The `Component` methods `get_shape` and `remove_shape` will need to be updated. I think subclassing `Component` for components with single-pixel shape descriptions and then overriding this method and making use of polymorphism is probably the best approach.

### LINKED ISSUE #447: Reduce number of NX classes tested in parameterised ui tests
A couple of the UI tests go through every possible NX class which means sitting through an extra 20 seconds of seizure-inducing tests. I think the list of classes to test should be reduced to around 5, making sure this includes any for which the interface is specialised (NXdetector etc).

matthew-d-jones:
### Issue
Closes #434  
Closes #447

### Description of work

Component.shape now optionally returns positions as well as the shape, if positions are returned then the 3D view creates a `QEntity` at each position.

### Acceptance Criteria 

Using the "Add component" interface, add an `NXdetector` with "single pixel shape".
Try from an OFF file and with cylinders.
Pixels should appear in the 3D view, and disappear if the component is deleted.


rerpha:
i get this when adding a chopper with no shape: 

```
Unable to create chopper geometry - Required field(s) missing: slit_height, slits, radius, slit_edges
Validation failed. Unable to create disk chopper mesh.
  File "/home/jack/code/repos/nexus-constructor/nexus_constructor/instrument_view.py", line 217, in add_component
    mesh = OffMesh(geometry.off_geometry)
AttributeError: 'NoneType' object has no attribute 'off_geometry'

Original exception was:
  File "/home/jack/code/repos/nexus-constructor/nexus_constructor/instrument_view.py", line 217, in add_component
    mesh = OffMesh(geometry.off_geometry)
AttributeError: 'NoneType' object has no attribute 'off_geometry'
```

nexus_constructor/instrument_view.py
```diff
@@ -201,18 +203,45 @@ def create_camera_filter(viewport, visible_entity, camera_to_filter):
...
+        for position in positions:
+            transform = Qt3DCore.QTransform()
+            transform.setTranslation(position)
+            position_transforms.append(transform)
+        self.component_positions[name] = position_transforms
+        # Note, the a list comprehension like below doesn't work (end up with segfault)
+        # self.component_positions[name] = [Qt3DCore.QTransform().setTranslation(position) for position in positions]
+        self.create_entities(name)
+
+    def create_entities(self, name):
```
rerpha: can this be called `_create_entities` as it's private?

nexus_constructor/instrument_view.py
```diff
@@ -226,8 +255,8 @@ def delete_component(self, name):
         Delete a component from the InstrumentView by removing the components and entity from the dictionaries.
         :param name: The name of the component.
         """
-
-        self.component_entities[name].setParent(None)
+        for entity in self.component_entities[name]:
+            entity.setParent(None)
```
rerpha: shouldn't this be done with the context manager anyway? could it be done with `del` instead?

tests/ui_tests/test_ui_add_component_window.py
```diff
@@ -65,6 +65,20 @@
     else:
         NO_PIXEL_OPTIONS[component_class] = i
 
+# Select a subset of the component class to use in parameterised tests
+# Should include any for which the UI is specialised
+_components_subset = {"NXdetector", "NXdisk_chopper", "NXsensor"}
```
rerpha: list literal instead of dict? both work, but lists are more probably more pythonic

nexus_constructor/component/component_shape.py
```diff
@@ -0,0 +1,49 @@
...
+
+SHAPE_GROUP_NAME = "shape"
+PIXEL_SHAPE_GROUP_NAME = "pixel_shape"
+CYLINDRICAL_GEOMETRY_NEXUS_NAME = "NXcylindrical_geometry"
+OFF_GEOMETRY_NEXUS_NAME = "NXoff_geometry"
+
+
+def get_shape_from_component(
+    component_group: h5py.Group, nexus_file: nx.NexusWrapper, shape_group_name: str
+):
```
rerpha: type hint for return type missing here unless it's causing circular imports

nexus_constructor/component/component_shape.py
```diff
@@ -0,0 +1,49 @@
...
+    # else return a placeholder to indicate the component's position
+    return NoShapeGeometry()
+
+
+class ComponentShape:
+    def __init__(self, nexus_file: nx.NexusWrapper, component_group: h5py.Group):
+        self.file = nexus_file
+        self.component_group = component_group
+
+    def get_shape(self):
```
rerpha: return type hint missing

nexus_constructor/component/component_type.py
```diff
@@ -2,7 +2,7 @@
 import os
 import xmltodict
 
-PIXEL_COMPONENT_TYPES = {"NXdetector", "NXdetector_module"}
+PIXEL_COMPONENT_TYPES = {"NXdetector"}
```
rerpha: [] instead of {}?

tests/ui_tests/test_ui_add_component_window.py
matthew-d-jones: It's a set, not a dict, seems most appropriate here to me.

nexus_constructor/instrument_view.py
matthew-d-jones: The context manager is to temporarily disconnect the entities. In this case we never want to reconnect them.
`del` does not do the same thing, if we delete the component before disconnecting it then we get segfaults.

nexus_constructor/component/component_type.py
matthew-d-jones: A `set` seems appropriate here to me

nexus_constructor/component/chopper_shape.py
```diff
@@ -0,0 +1,37 @@
...
+        chopper_checker = NexusDefinedChopperChecker(self.component_group)
+        if chopper_checker.validate_chopper():
+            return (
+                DiskChopperGeometryCreator(
+                    chopper_checker.chopper_details
+                ).create_disk_chopper_geometry(),
+                None,
+            )
+        else:
+            print("Validation failed. Unable to create disk chopper mesh.")
```
rerpha: we should remember to change this to a log statement if it is merged before #468

matthew-d-jones:
Merged as it has passed on CentOS. Mac and Windows nodes are still down.

Issue #468, opened 2019-10-09
Use logging library


matthew-d-jones:
https://docs.python.org/3.6/library/logging.html

Convert all current `print` statements to use logging library.

Issue #413, opened 2019-07-23
Remove red cube if there is no sample or if the sample has a shape


rerpha:
Currently even if a sample is removed and a file is loaded in the red cube which indicates a sample is still displayed. It is hardcoded in and has nothing to do with the model, which is an issue. It should be related to the default sample in the h5py file and if removed should not be displayed.

rerpha:
Check this is done before closing

matthew-d-jones:
This is working. It retains the red, low opacity, material and renders whatever the shape of the sample is.

PR #474, opened 2019-10-09
Only removing components from the 3d view if they have a shape. 

### LINKED ISSUE #395: KeyError when deleting components from treeview
# How to replicate: 
- press duplicate loads of times on the a component 
- delete several of the duplicates 

See errors reported:
```
ERROR:root:Unable to delete component sample4 because it doesn't exist.
```
Doesn't seem to be affected by whether the component has a shape or not.

rerpha:
### Issue

Closes #395 

### Description of work
This just fixes an issue with the sample cube not being attached to the sample in the nexus file - if you duplicate or delete it then it causes an exception usually. This just adds a check to see if there is an entity attached to a component before deleting the component entity. If not, then there is a statement below which deals with it and logs that the "component does not exist" - This has been changed to a log statement in #468 

### Acceptance Criteria 

Added an if statement

### UI tests

N/A - 3d view not testable due to segfault errors. 


rerpha:
think we decided we don't need this as it causes other errors, and really the sample needs to be attached to the shape of the sample, which would prevent this error from happening in the first place. Closing for now 

For more info see #413

PR #479, opened 2019-10-10
fix executable and windows build

rerpha:
### Issue

Closes #187 

### Description of work

Unpinning cx_freeze - the issue on windows has been fixed in 6.0
https://github.com/anthony-tuininga/cx_Freeze/issues/407

Also fixing the executable (had been broken for a while) and developing a test to make sure it doesn't break again. 

code changes mean that the definitions dir only has to be found once rather than all over the code.

### Acceptance Criteria 

N/A

### UI tests

N/A

### Nominate for Group Code Review

- [ ] Nominate for code review

PR #486, opened 2019-10-22
fix for windows - icon directory separator

rerpha:
### Issue

None

### Description of work

Fixes the nexus-constructor on windows - previously it would just crash due to a `QIcon` call trying to find an item that does not exist. 

### Acceptance Criteria 

It works on windows now. 

### UI tests

N/A


rerpha:
going to do this in #479

PR #490, opened 2019-10-23
489 correctly set parents

### LINKED ISSUE #489: Correctly set parent entities so that we do not need a python reference
Within the instrument view we have a lot of code that does for example:
```
def __init__():
    self.material = None

def create_object():
    self.material = Qt3DExtras.QPhongAlphaMaterial()
    create_entity(self.material)
```
This is because if a `QNode` does not have a parent it will be destroyed if it goes out of scope. By providing a valid parent when we create the QNode we need not hold a reference to the object within python if we're never going to use it again. Go through the GUI and refactor everywhere where this happens.

DominicOram:
### Issue

Closes #489

### Description of work

See ticket, refactored so that we needn't hold on to so many references.

### Acceptance Criteria 

* Confirm that everything still works, especially:
   * The gnomon, axes lines and sample still display as before
   * You can add a component with a mesh and it correctly displays
   * You can remove a component and it disappears

### UI tests

Behaviour is unchanged

### DIFF
--- a/nexus_constructor/instrument_view.py
+++ b/nexus_constructor/instrument_view.py
@@ -83,15 +83,10 @@ class InstrumentView(QWidget):
         self.surface_selector.setSurface(self.view)
 
         # Create the gnomon resources and get its camera
-        self.gnomon = Gnomon(
-            self.gnomon_root_entity,
-            self.view.camera(),
-        )
+        self.gnomon = Gnomon(self.gnomon_root_entity, self.view.camera())
 
         # Create the axes lines objects
-        InstrumentViewAxes(
-            self.axes_root_entity, self.view.camera().farPlane()
-        )
+        InstrumentViewAxes(self.axes_root_entity, self.view.camera().farPlane())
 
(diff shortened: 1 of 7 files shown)


nexus_constructor/instrument_view.py
```diff
@@ -218,40 +195,29 @@ def add_component(
...
+                transform = Qt3DCore.QTransform(self.component_root_entity)
+                transform.setTranslation(position)
+
+                material = create_material(
+                    QColor("black"), QColor("grey"), self.component_root_entity
                 )
-                for position_transform in self.component_positions[name]
-            ]
+
+                self.component_entities[name] = [
```
matthew-d-jones: This isn't quite right, now you overwrite `self.component_entities[name]` with a list contain a single entity each time.
The result is that we only have a handle for the last pixel that was created, so when we try to delete a detector bank component only that pixel is removed from the 3D view.

DominicOram: Good spot, thanks.

PR #491, opened 2019-10-24
481-  fix stream and link writing

### LINKED ISSUE #481: Bug in outputting streams and links to File Writer JSON
We store a `stream` field as a group in the in-memory NeXus file, but it should only result in a `stream` object in the JSON output, not a group containing a stream. In other words, there is an extra nested group in the JSON output which should not be there.

I think part of the confusion is that there is an edit box for a name when you add a stream to the field dialog. I don't think that should be there, instead we should make the field identifiable by displaying the topic (and maybe source if that exists) as a text label instead.

The `name` field of link objects in the JSON should be the name, not the full path from the root of the file.

rerpha:
### Issue

Closes #481 

### Description of work

Fixes the links and streams being put in an extra group that should not be there. This is now not the case and it just puts in the stream object. 

### Acceptance Criteria 

When "kafka stream" is selected as field type, the name field should not be visible. 

Streams and links should now be outputted properly 

### UI tests

N/A - no UI changes besides the kafka stream field disappearing. 

### DIFF
--- a/nexus_constructor/component_fields.py
+++ b/nexus_constructor/component_fields.py
@@ -73,8 +73,14 @@ class FieldWidget(QFrame):
         possible_field_names: List[str],
         parent: QListWidget = None,
         instrument: Instrument = None,
+        stream_group_name: str = None,
     ):
         super(FieldWidget, self).__init__(parent)
+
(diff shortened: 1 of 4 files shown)


rerpha:
I think merging in switch_to_widgets has fixed this?

PR #493, opened 2019-10-24
477 - deleting transforms from chain when components are deleted

### LINKED ISSUE #477: Properly handle the deletion of components with transformations in them
Deleting a component that has transformations in the UI currently results in several exceptions.

The code that fixes this needs to iterate over the transformations in the component and delete them one by one before deleting the actual component.

Whoever implements this should start doing so in the function `ComponentTreeModel.__remove_component()` in the file _component_tree_model.py_.

rerpha:
### Issue

Closes #477 

### Description of work

Deletes the transformations before deleting a component, which in-turn should restructure the depends_on tree. 

### Acceptance Criteria 

UI looks correct 
No exceptions thrown 
depends_on is correct
transformations are deleted from nexus file

### UI tests

None yet - want to check i am on the right path first. 

### DIFF
--- a/nexus_constructor/component_tree_model.py
+++ b/nexus_constructor/component_tree_model.py
@@ -124,13 +124,8 @@ class ComponentTreeModel(QAbstractItemModel):
 
     def __remove_component(self, index: QModelIndex):
         component = index.internalPointer()
-        is_depended_on = False
         transforms = component.transforms
-        if transforms:
-            if transforms[0].get_dependents():
-                is_depended_on = True
-
-        if is_depended_on:
+        if transforms and transforms[0].get_dependents():
             reply = QMessageBox.question(
                 None,
                 "Delete component?",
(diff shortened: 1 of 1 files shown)


rerpha:
TODO: add a dialog before deleting components that are dependees of other components to warn the user that they are about to do so.

rerpha:
> I did some simple tests and it still fails (produces an exception) when the component that you delete has a link to another component.
> The error message is in that case:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/jonasnilsson/Documents/nexus-constructor/nexus_constructor/instrument_view.py", line 230, in delete_component
>     for entity in self.component_entities[name]:
> KeyError: 'sample'
> ```

that's to do with the sample not being attached to the shape in the 3d view, separate to this issue. #395 is for that one.

SkyToGround:
Ok then, good job.

PR #513, opened 2019-11-05
Remove empty attributes list from json

### LINKED ISSUE #504: Avoid including empty attributes fields in output JSON
`"attributes": []` is unnecessary in the JSON output.

rerpha:
### Issue

Closes #504 

### Description of work
Removes the `attributes` list if there are none in the JSON. 

### Acceptance Criteria 

If something in the nexus file has no attributes, the `attributes` for this should not appear in the filewriter JSON. 

### UI tests

No UI changes.

Unit tests added for functionality. 

### DIFF
--- a/nexus_constructor/json/filewriter_json_writer.py
+++ b/nexus_constructor/json/filewriter_json_writer.py
@@ -67,12 +67,14 @@ def cast_to_int(data):
 
 
 def _add_attributes(root: NexusObject, root_dict: dict):
-    root_dict["attributes"] = []
+    attrs = []
     for attr_name, attr in root.attrs.items():
         if isinstance(attr, bytes):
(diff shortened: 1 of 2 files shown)


rerpha:
> Looks good to me. My only comment is that using a fixture for the `test_file` might make things a little neater in the long-term? I think if you `yield` the file in a fixture then you can add code that will close the file afterwards.

I think this is a good idea, essentially what that does is provides a tear-down functionality so even if the test fails it still closes the file. Will have a go at this now

rerpha:
how does that look @DolicaAkelloEgwel ?

DolicaAkelloEgwel:
> how does that look @DolicaAkelloEgwel ?
👍

PR #522, opened 2019-11-06
430: Edit Pixel Data

### LINKED ISSUE #430: Pixel Controls: Enable editing a component with PixelData
Putting this in a separate issue. I am also wondering what should happen if a pixel grid is replace with a pixel mapping and vice versa, or if the pixel info is removed altogether.

DolicaAkelloEgwel:
### Issue

Closes #430

### Description of work

- Enables editing/deleting/creating pixel data in an existing component. 
- `cylinders` in `CylindricalGeometry` is now a property.
- Half dealt with the case of one row/column, but this has not been fully sorted out. Creating a grid with a single row/column will probably lead to exceptions. I decided to set this aside for #531.

### Acceptance Criteria 

- Create a pixel grid. Increase and decrease the number of rows/columns (though 1 won't work). Change the row height/column width, the count first from value, the count direction, and the first ID.
- Create a pixel mapping. Change the pixel IDs.
- Create a component with a grid then edit the component and switch to mapping. 
- Create a component with a mapping then edit the component and switch to a grid.
- Create a component with a grid and then edit it and switch to no pixels.
- Create a component with a mapping and then edit it and switch to no pixels.
- Create a component with no pixels and then edit it and switch to a grid
- Create a component with no pixels and then edit it and switch to a mapping.

In all cases check that:
- The updated information is what appears in the dialog fields when you attempt to edit the component a second time.
- The changes are reflected in the NeXus File Layout.
- Any fields that are no longer needed are deleted from the NeXus file (e.g. `x_pixel_offset` is removed if you switch from grid to mapping/no pixels).
- The expected changes take place in the Instrument View.

Also creating a component with a pixel grid of "non-existent" cylinders (radius and length of zero) and then attempting to edit the component results in an error. Will open a separate issue for this.

### UI tests

n/a.


nexus_constructor/pixel_mapping_widget.py
```diff
@@ -28,6 +28,9 @@ def __init__(self, parent: QListWidget, id_no: int, text: str):
 
         self.setLayout(self.layout)
 
+    def set_id(self, pixel_id: int):
```
rerpha: can you use @property here?

nexus_constructor/pixel_options.py
```diff
@@ -10,6 +14,28 @@
...
+
+def check_data_is_an_array(data) -> bool:
+    """
+    At the moment it appears as though a scalar can still be returned as an array when using `get_field_value` (though
+    it could just be me doing the wrong thing). This function checks if an array contains more than one value so that
+    Pixel Data can be edited in the case of a Single Shape.
+    :param data: The data value from the NeXus file.
+    :return: True if the data is a scalar or an array containing a single value, False otherwise.
+    """
+    if type(data) is not np.ndarray:
```
rerpha: can use isinstance(data, np.ndarray) here

rerpha: or alternatively np.isscalar(data)

nexus_constructor/pixel_options.py
```diff
@@ -62,12 +88,142 @@ def setupUi(self, parent_widget):
...
+        first_id_plus_one = first_id + 1
+
+        count_along_text = COLUMNS_TEXT
+        right_of_first_id = (first_id_index[0], first_id_index[1] + 1)
+        left_of_first_id = (first_id_index[0], first_id_index[1] - 1)
+
+        start_counting_text = None
+
+        if first_id_index == (0, 0):
+            start_counting_text = TOP_LEFT_TEXT
```
rerpha: i think you can refactor this and the following lines out, and give TOP_LEFT_TEXT, right_of_first_id as arguments. besides that they are duplicated minus the conditional above it

nexus_constructor/pixel_options.py
```diff
@@ -62,12 +88,142 @@ def setupUi(self, parent_widget):
...
+            if detector_numbers[left_of_first_id] == first_id_plus_one:
+                count_along_text = ROWS_TEXT
+
+        return first_id, start_counting_text, count_along_text
+
+    def _fill_entire_shape_fields(self, component_to_edit: Component):
+
+        shape = component_to_edit.shape[0]
+
+        if type(shape) is OFFGeometryNexus:
```
rerpha: isinstance()

nexus_constructor/pixel_options.py
```diff
@@ -62,12 +88,142 @@ def setupUi(self, parent_widget):
...
+            self.create_pixel_mapping_list(n_faces, "face")
+
+            for detector_face in detector_faces:
+                self.pixel_mapping_widgets[detector_face[0]].set_id(detector_face[1])
+
+        else:
+
+            detector_number = shape.detector_number
+
+            if type(detector_number) is List:
```
rerpha: isinstance()

nexus_constructor/array_dataset_table_widget.py
```diff
@@ -125,6 +125,7 @@ def columnCount(self, parent: QModelIndex = ...) -> int:
 
     def data(self, index: QModelIndex, role: int = ...) -> str:
         if role == Qt.DisplayRole or role == Qt.EditRole:
+            print(self.array)
```
rerpha: is this needed?

nexus_constructor/invalid_field_names.py
```diff
@@ -2,4 +2,13 @@
...
+INVALID_FIELD_NAMES = [
+    "description",
+    "shape",
+    "depends_on",
+    "x_pixel_offset",
+    "y_pixel_offset",
+    "z_pixel_offset",
+    "detector_number",
+    "detector_faces",
+]
```
rerpha: :+1:

nexus_constructor/pixel_data_to_nexus_utils.py
```diff
@@ -66,17 +75,40 @@ def get_detector_ids_from_pixel_grid(grid: PixelGrid):
...
+    return convert_to_scalar_if_array_has_one_element(ids)
+
+
+def convert_to_scalar_if_array_has_one_element(value_array: np.ndarray):
+    """
+    Prevents a single value from being returned as an array.
+    :param value_array: The array containing pixel information.
+    :return: An array if the pixel information has more than 1 element, or a single value if the array has a size of 1.
+    """
+    if value_array.size > 1:
```
rerpha: Doesn't really matter, and i'm not sure this is even cleaner, but you could make the below function redundant by doing something like his: 
`if (value_array.size if isinstance(value_array, np.ndarray) else len(value_array)) > 1`

you could then use value_array[0] for both types

DolicaAkelloEgwel:
Kind of unsure what to do when a Pixel Grid has one row and/or one column. Can a pixel grid that consists of a single row/column have a row height/column width? At the moment I'm using the distance between the neighbouring pixels to compute these things. I'm not sure what to do in the case when no neighbouring pixel exists. For the time being I'm going to treat the row height/column width as being None for Pixel Grids that only contain a single row/column but I don't know if that's the best approach.

Thoughts?
@rerpha 
@SkyToGround

SkyToGround:
> Kind of unsure what to do when a Pixel Grid has one row and/or one column. Can a pixel grid that consists of a single row/column have a row height/column width? At the moment I'm using the distance between the neighbouring pixels to compute these things. I'm not sure what to do in the case when no neighbouring pixel exists. For the time being I'm going to treat the row height/column width as being None for Pixel Grids that only contain a single row/column but I don't know if that's the best approach.

I talked to the Mantid group about this and they did not know the best solution for this either. In the end we decided that they want to test at least two alternatives.
1. The width is set to 0.
2. The width is left out (not populated).

@DolicaAkelloEgwel If you can put together two simple NeXus files with those two alternatives then we can figure our which alternative is the best.

nexus_constructor/add_component_window.py
```diff
@@ -583,17 +600,19 @@ def populate_pixel_mapping_if_necessary(self):
...
 
-        if self.CylinderRadioButton.isChecked():
-            self.pixel_options.populate_pixel_mapping_list_with_cylinder_number(
-                self.cylinderCountSpinBox.value()
-            )
+            if self.CylinderRadioButton.isChecked():
+                self.pixel_options.populate_pixel_mapping_list_with_cylinder_number(
+                    self.cylinderCountSpinBox.value()
+                )
+        except RuntimeError:
```
DolicaAkelloEgwel: Not the best way to get around this...

tests/ui_tests/test_ui_pixel_options_widget.py
```diff
@@ -439,3 +586,256 @@ def test_UI_GIVEN_no_pixels_button_is_pressed_WHEN_entering_pixel_data_THEN_call
...
+    n_cylinders = cylindrical_geometry.cylinders.size / 3
+    assert n_cylinders == len(pixel_options.pixel_mapping_widgets)
+    assert (
+        pixel_options.pixel_mapping_widgets[0].id
+        == pixel_mapping_with_single_pixel.pixel_ids[0]
+    )
+
+
+@pytest.mark.parametrize("old_value", [5, 15])
+@pytest.mark.parametrize("new_value", [5, 15])
```
matthew-d-jones: I don't understand what this is testing, are old and new value supposed to be the same?

tests/ui_tests/test_ui_pixel_options_widget.py
```diff
@@ -439,3 +586,256 @@ def test_UI_GIVEN_no_pixels_button_is_pressed_WHEN_entering_pixel_data_THEN_call
...
+
+def test_GIVEN_array_with_single_element_WHEN_calling_check_data_is_an_array_THEN_returns_false():
+
+    data = np.array([3.5])
+    assert not check_data_is_an_array(data)
+
+
+def test_GIVEN_array_with_multiple_elements_WHEN_calling_check_data_is_an_array_THEN_returns_true():
+
+    data = np.array([i for i in range(5)])
```
matthew-d-jones: You can use numpy's version of the range function: `data = np.arange(5)`

tests/ui_tests/test_ui_pixel_options_widget.py
```diff
@@ -439,3 +586,256 @@ def test_UI_GIVEN_no_pixels_button_is_pressed_WHEN_entering_pixel_data_THEN_call
...
+    pixel_grid.columns = 1
+    x_pixel_offsets = get_x_offsets_from_pixel_grid(pixel_grid)
+
+    n_columns, column_width = pixel_options._get_column_information(x_pixel_offsets)
+
+    assert n_columns == pixel_grid.columns
+    assert column_width is None
+
+
+@pytest.mark.xfail
```
matthew-d-jones: Why is this test expected to fail?

PR #523, opened 2019-11-06
436 - Adding interface for setting attributes on fields

### LINKED ISSUE #436: Add interface for setting attributes on fields
Fields often have attributes, a particularly common example is the `units` attribute. Currently the user has no way of setting these.

rerpha:
### Issue

Closes #436 

### Description of work

Adds an interface so users can edit field attributes. This is done with a similar method to how fields are edited, just with scalar and arrays only. 

Input should be validated in the same way fields are, so entering "hello" with a scalar field that has a double dtype should not work. 

Also makes field.value return either a softlink, group or dataset for consistency and ease of use. 

### Acceptance Criteria 

Attrs should now be added when they are created through fields - this should be applied to the nexus and json files when exporting. 

### UI tests

Validators already covered 


rerpha:
I can't figure out why the UI test is failing, might need some help on this one.

PR #525, opened 2019-11-11
Add file writer control to nexus-constructor

SkyToGround:
### Description of work

In order to simplify my own job I have added support for controlling the file-writer directly from the nexus-constructor. Please comment on the existence and implementation of this code.


nexus_constructor/main_window.py
```diff
@@ -169,6 +171,32 @@ def show_edit_component_dialog(self):
         ].internalPointer()
         self.show_add_component_window(selected_component)
 
+    def _setupFileWriterCtrl(self, main_window):
+        try:
+            import kafka  # noqa: F401
+
+            self.control_file_writer_action = QtWidgets.QAction(main_window)
```
rerpha: Can this be moved to main_window.py in `ui/`? i also don't think this is getting added to the main layout, so at the moment won't be visible.

nexus_constructor/main_window.py
```diff
@@ -169,6 +171,32 @@ def show_edit_component_dialog(self):
...
+            self.file_menu.addAction(self.control_file_writer_action)
+            self.control_file_writer_action.setText(
+                QtWidgets.QApplication.translate(
+                    "MainWindow", "Control file-writer", None, -1
+                )
+            )
+            self.control_file_writer_action.triggered.connect(
+                self.show_control_file_writer_window
+            )
+        except ImportError:
```
rerpha: we will then not need any of this stuff

nexus_constructor/main_window.py
```diff
@@ -169,6 +171,32 @@ def show_edit_component_dialog(self):
...
+            pass
+
+    def show_control_file_writer_window(self):
+        if self.file_writer_ctrl_window is None:
+            from nexus_constructor.file_writer_ctrl_window import FileWriterCtrl
+
+            self.file_writer_ctrl_window = FileWriterCtrl(self.instrument)
```
rerpha: This should stay in here in my opinion.

nexus_constructor/file_writer_ctrl_window.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+    Sets the line edit colour to red if field is invalid or white if valid. Also sets the tooltips if provided.
+    :param line_edit: The line edit object to apply the validation to.
+    :param is_valid: Whether the line edit field contains valid text
+    :param suggestion_callable: A callable that returns the suggested alternative if not valid.
+    :param tooltip_on_accept: Tooltip to display if line edit is valid.
+    :param tooltip_on_reject: Tooltip to display if line edit is invalid.
+    :return: None.
+    """
+    colour = "white" if is_valid else "red"
```
rerpha: this method already exists in ui_utils.py. Can you use that method instead?

nexus_constructor/file_writer_ctrl_window.py
```diff
@@ -0,0 +1,333 @@
...
+        self.commandBrokerLed.turn_on(False)
+        self.commandBrokerEdit.textChanged.connect(self.onCommandBrokerTextChanged)
+        self.commandBrokerChangeTimer = QTimer()
+        self.commandBrokerChangeTimer.setSingleShot(True)
+        self.commandBrokerChangeTimer.timeout.connect(
+            self.onCommandBrokerTextChangeTimer
+        )
+        self.command_producer = None
+
+        self.fileNameLineEdit.textChanged.connect(self.onFileNameChange)
```
rerpha: I am wondering if some of the code in filewriter_command_dialog can be used, possibly even the whole thing? it seems like it could at least be inherited from, seeing as it's doing the same kind of thing?

ui/filewriter_ctrl_frame.py
```diff
@@ -0,0 +1,199 @@
+# -*- coding: utf-8 -*-
+
+# Form implementation generated from reading ui file 'filewriter_command_frame.ui',
+# licensing of 'filewriter_command_frame.ui' applies.
+#
+# Created: Mon Nov 11 16:21:42 2019
+#      by: pyside2-uic  running on PySide2 5.13.0
+#
```
rerpha: may as well remove this bit at the top as we don't have the UI file

ui/filewriter_ctrl_frame.py
```diff
@@ -0,0 +1,199 @@
...
+# WARNING! All changes made in this file will be lost!
+
+from PySide2 import QtCore, QtGui, QtWidgets
+
+class Ui_FilewriterCtrl(object):
+    def setupUi(self, FilewriterCtrl):
+        FilewriterCtrl.setObjectName("FilewriterCtrl")
```
rerpha: i think this could do with refactoring into simple methods that are responsible for setting up certain parts of the UI.

nexus_constructor/StatusConsumer.py
```diff
@@ -0,0 +1,111 @@
...
+        self._setConnected(True)
+        known_writers = {}
+        known_files = {}
+        while self.run_thread:
+            data = consumer.poll(500)
+            had_updates = False
+            for itm in data:
+                had_updates = True
+                for msg in data[itm]:
+                    msg_obj = json.loads(msg.value)
```
rerpha: i think this should be refactored out, it is very nested. From here onwards it could just be `_handle_status_message()`

nexus_constructor/StatusConsumer.py
```diff
@@ -0,0 +1,111 @@
...
+        return return_filewriters
+
+    def getFiles(self):
+        self.status_lock.acquire()
+        return_files = copy(self.files)
+        self.status_lock.release()
+        return return_files
+
+    def _setConnected(self, is_connected):
+        self.status_lock.acquire()
```
rerpha: maybe a bit nit-picky but could this be done with a context manager of some sorts? it is duplicated in most functions

nexus_constructor/CommandProducer.py
```diff
@@ -0,0 +1,65 @@
...
+    def hasUnsentMessages(self):
+        return not self.msg_queue.empty()
+
+    def isConnected(self):
+        self.command_lock.acquire()
+        return_status = copy(self.connected)
+        self.command_lock.release()
+        return return_status
+
+    def sendCommand(self, message):
```
rerpha: type hint missing

nexus_constructor/CommandProducer.py
```diff
@@ -0,0 +1,65 @@
...
+        self.msg_queue.put(message, block=True)
+
+    def _setConnected(self, is_connected):
```
rerpha: type hint missing

nexus_constructor/CommandProducer.py
```diff
@@ -0,0 +1,65 @@
+from kafka import KafkaProducer
+from kafka.errors import NoBrokersAvailable
+import threading
+import time
+from queue import Queue
+from copy import copy
+
+
+class CommandProducer:
+    def __init__(self, address, topic):
```
rerpha: type hints missing

nexus_constructor/StatusConsumer.py
```diff
@@ -0,0 +1,111 @@
+from kafka import KafkaConsumer, TopicPartition
+import threading
+from copy import copy
+import time
+from kafka.errors import NoBrokersAvailable
+import json
+
+
+class StatusConsumer:
+    def __init__(self, broker, topic):
```
rerpha: type hints

nexus_constructor/StatusConsumer.py
```diff
@@ -0,0 +1,111 @@
...
+        self.run_thread = False
+        self.thread.join()
+
+    def isConnected(self):
+        self.status_lock.acquire()
+        return_status = copy(self.connected)
+        self.status_lock.release()
+        return return_status
+
+    def getFilewriters(self):
```
rerpha: could use `@property` here along with `@property.setter`? for each of these functions

ui/filewriter_ctrl_frame.py
```diff
@@ -0,0 +1,199 @@
...
+        self.brokerLineEdit.setObjectName("brokerLineEdit")
+        self.commandFormLayout.setWidget(1, QtWidgets.QFormLayout.FieldRole, self.brokerLineEdit)
+        self.brokerLabel = QtWidgets.QLabel(self.centralwidget)
+        self.brokerLabel.setObjectName("brokerLabel")
+        self.commandFormLayout.setWidget(1, QtWidgets.QFormLayout.LabelRole, self.brokerLabel)
+        self.startTimeLayout = QtWidgets.QHBoxLayout()
+        self.startTimeLayout.setObjectName("startTimeLayout")
+        self.useStartTimeCheckBox = QtWidgets.QCheckBox(self.centralwidget)
+        self.useStartTimeCheckBox.setText("")
+        self.useStartTimeCheckBox.setObjectName("useStartTimeCheckBox")
```
rerpha: I'm not sure these setObjectName calls are strictly necessary - they are only used when trying to find a global QObject. If they are not necessary, may as well remove them.

rerpha:
I think this is a good idea overall and seeing as we are already able to produce the messages we may as well send them. I wonder if we should also be able to send forwarder messages too? 

also, if this is going to be in production, we will need to add kafka-python to `requirements.txt`

PR #560, opened 2019-11-28
V20 FIXES

### LINKED ISSUE #559: stream advanced options that are 0 should not be added
currently all of the advanced options are getting added even if they are set to their default values. This should be prevented.

### LINKED ISSUE #562: Streams that are loaded from JSON are not put in the correct place
Fixed in #560 

they end up in a group with the source as the name - they should have the name of the parent directory the source name group currently sits in.

### LINKED ISSUE #565: h5py deprecation warning: The default file mode will change to 'r' (read-only) in h5py 3.0.
we use h5py.File in a few places without a write mode so we need to add modes to these to suppress these warnings or we will end up with read-only in-memory files.

### LINKED ISSUE #568: Remove topic line edit placeholder text
it is misleading as it suggests you need to specify the broker but this is not the case.

### LINKED ISSUE #569: add a way to specify a default broker when writing forwarder commands
currently there is no way to get the broker information into the forwarder command like there is in the filewriter command.

rerpha:
### Issue

Closes #562 
Closes #568
Closes #559 
Closes #565 
Closes #569 

### Description of work

I have logged issues while at V20 and fixed them on this branch. 

### Acceptance Criteria 

It works at V20 

### UI tests

None changed yet. 

### DIFF
--- a/nexus_constructor/add_component_window.py
+++ b/nexus_constructor/add_component_window.py
@@ -97,6 +97,7 @@ def update_existing_stream_field(field: h5py.Dataset, new_ui_field: FieldWidget)
     :param field: The dataset to copy into the value line edit
     :param new_ui_field: The new UI field to fill in with existing data
     """
+    new_ui_field.name = get_name_of_node(field)
     new_ui_field.field_type = FieldType.kafka_stream.value
     new_ui_field.streams_widget.update_existing_stream_info(field)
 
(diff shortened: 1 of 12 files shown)


matthew-d-jones:
You haven't created a separate PR for the fix for #562, is that intentional?

rerpha:
> 
> 
> You haven't created a separate PR for the fix for #562, is that intentional?

nah, just haven't got round to it yet.

rerpha:
Closing, have created PR for all of the issues fixed in this PR.

PR #577, opened 2019-12-04
540: Make disk chopper with attributes

### LINKED ISSUE #540: Use attributes interface to enter units for disk choppers
Metres and degrees are used by default in the `UserDefinedChopperChecker`. This can be changed now that there's an interface for entering attributes. 

Also just noticed that the second check in the `_input_describes_valid_chopper` method assumes the `slit_height` and `radius` have the same units. So now that units can be entered, that would have to be changed to prevent it from thinking a radius of 3cm is "smaller" than a slit height of 14mm.

### LINKED ISSUE #580: Can't seem to edit components with 1D array fields
```
  File "/home/dolica/nexus-constructor/nexus_constructor/array_dataset_table_widget.py", line 128, in data
    value = self.array[index.row()][index.column()]
IndexError: invalid index to scalar variable.
```

DolicaAkelloEgwel:
### Issue

Closes #540
Closes #580 

### Description of work

- Allows you to set the units for the radius, slit height, and slit edges of a disk chopper using attributes in the Add Component Window. The Disk Chopper should also be completely editable.
- Also added some tests for the `ArrayDatasetTableWidget` in the process of figuring out #580.
- The functions for validating units have now been moved to their own file so that they can be shared with the `UnitValidator` and the `NexusDefinedChopperChecker`.
- Units input for the chopper attributes should also be able to handle metres, m, rad, radians, etc due to `pint`.
- Also tried to improve a few logging messages for the chopper validation.
- `UserDefinedChopperChecker` is no more.

### Acceptance Criteria 

Create a disk chopper and set units then see that it works. Check that the tests pass. Edit a disk chopper by changing its radius/units/etc and see that that works, and that the previous chopper is removed from the view.

### UI tests

n/a


DominicOram:
Waiting as will be easier to review after https://github.com/ess-dmsc/nexus-constructor/issues/579.

tests/ui_tests/test_ui_array_dataset_table_widget.py
```diff
@@ -0,0 +1,154 @@
...
+    array_dataset_table_widget.model.update_array_dtype(new_data_type)
+
+    assert np.array_equal(array, array_dataset_table_widget.model.array)
+
+
+@pytest.mark.skip("Don't actually know what causes the ValueError.")
+def test_UI_GIVEN_array_cant_be_converted_WHEN_changing_data_type_THEN_array_resets(
+    array_dataset_table_widget
+):
+    pass
```
DominicOram: Please implement

nexus_constructor/geometry/disk_chopper/disk_chopper_checker.py
```diff
@@ -6,7 +6,11 @@
 from h5py import Group
```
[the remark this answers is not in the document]
DolicaAkelloEgwel: Four out of the five functions used for doing checks are common functions. Or are you talking about something else?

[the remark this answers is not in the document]
DominicOram: I was thinking `validate_chopper`, `chopper_details` and the constructors are similar enough to share but it's probably outside the scope of this ticket, don't worry.

nexus_constructor/array_dataset_table_widget.py
```diff
@@ -125,7 +125,9 @@ def columnCount(self, parent: QModelIndex = ...) -> int:
 
     def data(self, index: QModelIndex, role: int = ...) -> str:
         if role == Qt.DisplayRole or role == Qt.EditRole:
-            value = self.array[index.row()][index.column()]
+            value = self.array[index.row()]
+            if isinstance(value, list) or isinstance(value, np.ndarray):
```
rerpha: can you use `not np.isscalar()` here?

nexus_constructor/unit_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    :return: True if the unit is contained in the pint registry, False otherwise.
+    """
+    try:
+        ureg(input)
+    except (
+        pint.errors.UndefinedUnitError,
+        AttributeError,
+        pint.compat.tokenize.TokenError,
+    ):
+        return False
```
rerpha: log to info/debug here

nexus_constructor/unit_utils.py
```diff
@@ -0,0 +1,58 @@
...
+def units_are_expected_type(input: str, expected_unit_type: str) -> bool:
+    """
+    Checks if a unit is the expected type by trying to convert it.
+    :param input: The units string.
+    :return: True if the conversion was successful, False otherwise.
+    """
+    try:
+        ureg(input).to(expected_unit_type)
+    except (pint.errors.DimensionalityError, ValueError, AttributeError):
+        return False
```
rerpha: log to info/debug here

Issue #550, opened 2019-11-27
Remove chunk settings code (and related cruft)


SkyToGround:
We are re-writing the file-writer to not use nested structures for its component settings. For example for setting the chunk size and and log interval. Hence we need to remove this part of the code from the nexus-constructor as well. This should also fix a few bugs for us.

rerpha:
Marking as blocked as this has not yet been implemented in the filewriter. Once it has been implemented and the documentation for the commands have been updated this can be done.

rerpha:
the code in #599 can be removed when this has been done

rerpha:
Is there a jira ticket @SkyToGround ? or has it already been done?

SkyToGround:
@rerpha I do know that there is a JIRA issue for fixing this in the file-writer but I do not remember the state of it. I do not remember the state of this in the nexus-constructor either. I also do not feel I have time to look at it right now. Hence, I suggest we leave this ticket as is and I will get back to this after we are done with the re-write of the file-writer. Unless you want to work on it of course.

rerpha:
Found it: [DM-1719](https://jira.esss.lu.se/browse/DM-1719)

rerpha:
Closing, the filewriter already uses dot notation instead of nested groups, the ticket is just for updating the fw docs to reflect this.

PR #605, opened 2019-12-19
Stopping NCstream from being added to JSON stream groups NX_class

### LINKED ISSUE #604: NCstream NX_class should never end up in filewriter JSON
Currently it is being added to the stream group in the JSON. Other attributes should be allowed but the writer module will not overwrite the nx_class if it is already present, so we need to remove it.

rerpha:
### Issue

Closes #604 

### Description of work
Refactored out some of the stuff that finds streams as there was some duplication going on and I think it should be more streamlined. Previously, you had to pass two dictionaries containing links and streams to the JSON writer. This was duplication, as these could be found from the root object passed to that method anyway. I have removed this and made it flow a bit nicer, and it should make the interface nicer to use. 

I have added a test that makes sure NCstream is not added to the nx_class attributes so this can be overwritten by the writer module. 

### Acceptance Criteria 

- Refactored `_handle_stream()` and `_handle_links()` in json writer 
- Removed `get_streams()` and `get_links()` from `instrument.py`
- Refactored finding forwarder streams into the forwarder JSON writer and hopefully making the code that does this a bit more readable 
- Hopefully made the code more easy to understand? 

### UI tests

None modified as no UI changes. 

### DIFF
--- a/nexus_constructor/instrument.py
+++ b/nexus_constructor/instrument.py
@@ -1,4 +1,4 @@
-from typing import List, Dict
+from typing import List
 
 import h5py
 from nexus_constructor.component.component_type import (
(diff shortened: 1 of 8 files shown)


rerpha:
still todo is remove the similar link-passing code. still in progress

rerpha:
I am not worried about the code coverage dropping btw - the dot hierarchy is being removed as per #550 anyway.

PR #606, opened 2019-12-20
#562: fix stream loading from JSON

### LINKED ISSUE #562: Streams that are loaded from JSON are not put in the correct place

rerpha:
### Issue

Closes #562 

### Description of work

Fixes stream groups being named incorrectly when loaded from JSON.

### Acceptance Criteria 

Fixed incorrect stream group name loading 
Improved test to check for stream fields

### UI tests

No UI changes. 


tests/test_json_writer.py
```diff
@@ -246,6 +246,7 @@ def test_GIVEN_stream_in_group_children_WHEN_handling_group_THEN_stream_is_appen
     converter = NexusToDictConverter()
     root_dict = converter.convert(file)
 
+    assert len(root_dict["children"]) == 1
```
matthew-d-jones: It would be good to explain why you are testing this. You can use `assert <condition>, "some string"`

Issue #567, opened 2019-11-29
executable should work from any location 


rerpha:
At V20 we had to cd into the build folder to get it to work correctly otherwise it complained that it could not find the definitions directory. It would be nice to be able to run it from anywhere, for example a desktop shortcut, without worrying about where the definitions directory is.

rerpha:
this is really difficult. CX_freeze appears to put stuff in weird locations, so doing something like `os.path.dirname(inspect.stack()[-1][1])` does not work as the stack keeps importing python. This means we can't get where the executable is running from, and therefore are unable to find the definitions dir. It works perfectly when used on the python file but not with the executable.

PR #609, opened 2020-01-13
575: link, stream, array, group for transformation magnitude

### LINKED ISSUE #575: Transformation magnitude should be allowed to be a Kafka stream, link, array etc.
Maybe refactor out the value part of the field widget into a separate widget and then reuse that in the transformation interface? There may be other places can be used too.

Related to DM-1789, requested by Tobias.

rerpha:
### Issue

Closes #575 
Unblocks #615 

### Description of work

Adds the ability to use field types as the transformation magnitude value. This means the transformation can be streamed from Kafka or linked to in the file. 

I also refactored out some of the transformation UI to make it simpler. 

### Acceptance Criteria 

Added a "ui_value" field in the file and then wrapped it with the transformations class. This should never end up in the json file and will still work when files are loaded in. 

Moved the file i/o that generates the nx class definitions so it is now done in one place instead of three - this wont really give any performance benefits but should make the code a bit more streamlined. 

### UI tests

Added some UI tests to cover the transformation view UI and underlying model. 
Added a couple of missing UI tests for the stream fields widget, as they are used in this PR anyway. 


rerpha:
Going to add a few more tests to the stream fields widget as this is now being used in two places and isn't fully covered.

rerpha:
Todo: add code and tests for dataset not being castable for scalar data

rerpha:
I think I will make a new ticket for the array slider as this PR is already quite big in terms of lines changed. everything else should now work, hopefully the UI has improved. 

The only thing is, the UI does not stay in-sync with the dataset value if it has a dtype that is not castable to a python `int`. I think this is correct behaviour?

rerpha:
have just realised i can easily fix #567 as well so going to push that change.

PR #616, opened 2020-01-22
611: component treeview refactoring from mainwindow

### LINKED ISSUE #553: Renaming a component with translations does not work
Renaming a component that is either depended on or depends on something else causes the tree view to fall over completely.

### LINKED ISSUE #611: Break out component treeview from main_window.py
The component treeview and its toolbar could be made into a composite widget, separate to the main window. This would further reduce the complexity of `nexus_constructor/main_window.py`.

rerpha:
### Issue

Closes #611 
Closes #553 

### Description of work

Breaks out the treeview UI from the main window. 
Also breaks up the code and hopefully makes it easier to understand

### Acceptance Criteria 

No functional changes

### UI tests

No functional changes - none modified. 


rerpha:
to do: fix renaming transformations 
to do: fix renaming components causing UI to have issues 

i think the model may need to be updated every time the nexus file is updated, and thus the view redrawn. Unless we make a component_changed signal that updates a single item in the view. I think it would be easier and less prone to errors if we just redrew the view each time anything is changed.

PR #622, opened 2020-01-24
re #621: relative path depends_on fix

### LINKED ISSUE #621: relative depends_on paths do not work
for example in the `bigfake.nxs` file in `sample/depends_on`

rerpha:
### Issue

Closes #621  

### Description of work

Fixes relative paths when generating dependee_of attribute for the UI to use. 
The `bigfake.nxs` file still has some issues loading but they are separate to this. This allows mostly everything to work in the UI. 

### Acceptance Criteria 

Added try/catch for relative + absolute paths in depends_on field
refactored method for setting the dependee_of attribute

### UI tests

None changed. 

### DIFF
--- a/nexus_constructor/nexus/nexus_wrapper.py
+++ b/nexus_constructor/nexus/nexus_wrapper.py
@@ -2,7 +2,7 @@ import logging
 import uuid
 import h5py
 from PySide2.QtCore import Signal, QObject
-from typing import Any, TypeVar, Optional, List
+from typing import Any, TypeVar, Optional
 import numpy as np
 
 h5Node = TypeVar("h5Node", h5py.Group, h5py.Dataset)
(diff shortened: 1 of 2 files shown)


nexus_constructor/nexus/nexus_wrapper.py
```diff
@@ -204,12 +204,21 @@ def find_depends_on_for_components(_, node):
...
             )
 
+        for transform, dependee_of in transforms_dependee_of.items():
+            try:
+                # numpy should cast to a scalar value if there is just one item.
+                # try and use an absolute path
+                set_dependee_of_attr(transform, dependee_of)
+            except KeyError:
+                # try and use a relative path instead
+                path = f"{dependee_of[0]}/{transform.decode('UTF-8')}"
```
matthew-d-jones: This can fail in the case that the transformation is a dependee of multiple components. The key of `transforms_dependee_of` _must_ already be the absolute path to the transform, otherwise it is possible to have multiple entries in the dictionary for the same transformation (because the same transformation can have different relative paths). I guess this would result in the dependee_of list getting overwritten and therefore missing entries.

Please add a unit test to cover this situation.
```
ComponentA
    depends_on: Transforms1/transform1
    Transforms1
        transform1
ComponentB
     depends_on: "/entry/instrument/ComponentA/Transforms1/transform1
```

PR #624, opened 2020-01-24
Refactor common attribute/field names into class to avoid typos

### LINKED ISSUE #618: replace dependee_of, ui_value with class containing them as constant vars
there are lots of places they are manually spelt out in a string which is prone to typos. We should replace any usage of dependee_of or ui_value to use the const vars instead. 

Also it might be worth putting NC as the attribute name prefix as they could cause issues if there is a file that contains attribtues with the same name.

rerpha:
### Issue

Closes #618 

### Description of work

Refactors out into const vars commonly used attribute and field names to avoid typos.
Also renames dependee_of and ui_value to include NC prefix as they are specific to the nexus constructor. 

### Acceptance Criteria 

No functional changes

### UI tests

None modified. 

### DIFF
--- a/nexus_constructor/field_utils.py
+++ b/nexus_constructor/field_utils.py
@@ -88,7 +88,7 @@ def find_field_type(item):
             return item, update_existing_link_field
         elif (
             CommonAttrs.NX_CLASS in item.attrs.keys()
-            and item.attrs[CommonAttrs.NX_CLASS] == "NCstream"
+            and item.attrs[CommonAttrs.NX_CLASS] == CommonAttrs.NC_STREAM
         ):
             return item, update_existing_stream_field
     logging.debug(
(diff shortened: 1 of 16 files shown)


nexus_constructor/field_utils.py
```diff
@@ -85,7 +86,10 @@ def find_field_type(item):
     elif isinstance(item, h5py.Group):
         if isinstance(item.parent.get(item.name, getlink=True), h5py.SoftLink):
             return item, update_existing_link_field
-        elif "NX_class" in item.attrs.keys() and item.attrs["NX_class"] == "NCstream":
+        elif (
+            CommonAttrs.NX_CLASS in item.attrs.keys()
+            and item.attrs[CommonAttrs.NX_CLASS] == "NCstream"
```
matthew-d-jones: Could add `NCstream` to `CommonAttrs`?

PR #635, opened 2020-02-18
Add filewriter control window

### LINKED ISSUE #600: Add live filewriter control
Jonas has already implemented this, we should add it as a CLI option

rerpha:
### Issue
Follows on from #525
Closes #600 

### Description of work

Just copied Jonas' changes to this PR, but I am going to refine the code a bit so we can get this merged if we decide it should be. 


rerpha:
- Add UI tests 
- Test any logic in filewriter control window that isn't UI-related

matthew-d-jones:
@rerpha I've updated the branch to use confluent-kafka and tested it against the file-writer v3.3.0 branch, so i'll throw this back to you now.
I noticed that the stop time control starts being disabled/enabled only after toggling the "stop time enabled" tickbox. Could you fix this so that it's initial state is disabled please?
At that point I think we can merge this branch if we don't find any problems.

PR #647, opened 2020-02-21
623 generate ui values for transforms

### LINKED ISSUE #623: ui_value is not generated when file is loaded
transformation datasets with no value will need to be given the ui_value attribute and set to 0. When loading the bigfake.nxs file it does not load some of the transformations due to the lack of the attribute.

rerpha:
### Issue

Closes #623 

### Description of work

Fixes all of the transformation views not showing in `bigfake.nxs` file as well as transformations not showing their UI placeholder values on file load. 

### Acceptance Criteria 

Turns out the UI problem is due to the spinbox not allowing negative values, so i chnaged the range to the other spinbox limits. 

### UI tests

None


nexus_constructor/transformations.py
```diff
@@ -157,14 +157,29 @@ def ui_value(self) -> float:
...
+                    return float(self._dataset[()])
+                except ValueError:
+                    logging.debug(
+                        "transformation value is not cast-able to float/int, using UI placeholder value instead."
+                    )
+            else:
+                # dataset value is array
+                try:
+                    self.ui_value = int(self._dataset[...][0])
+                    return int(self._dataset[...][0])
```
matthew-d-jones: why `int`?

nexus_constructor/transformations.py
```diff
@@ -157,14 +157,29 @@ def ui_value(self) -> float:
...
+            if np.isscalar(self.dataset[()]):
+                try:
+                    self.ui_value = float(self._dataset[()])
+                    return float(self._dataset[()])
+                except ValueError:
+                    logging.debug(
+                        "transformation value is not cast-able to float/int, using UI placeholder value instead."
+                    )
+            else:
+                # dataset value is array
```
matthew-d-jones: Would be good to explain in the comment, or ideally a unit test (possible?), that we'll use the first value in the array for 3D view display purposes.

nexus_constructor/transformations.py
rerpha: sorry, that should be float

nexus_constructor/transformations.py
rerpha: Have added to the comment and added a couple of unit tests to cover.

nexus_constructor/transformations.py
```diff
@@ -157,14 +157,29 @@ def ui_value(self) -> float:
...
+                    logging.debug(
+                        "transformation value is not cast-able to float/int, using UI placeholder value instead."
+                    )
+            else:
+                # Dataset value is array - use the first value of the array as the UI value
+                try:
+                    self.ui_value = float(self._dataset[...][0])
+                    return float(self._dataset[...][0])
+                except ValueError:
+                    # not int-type
```
matthew-d-jones: comment incorrect, should be something like "not castable to float" maybe?
Rather than `pass` should it return a default value (0.0)?
Could have a `default_value` variable and use that also on lines 180 and 181.

PR #675, opened 2020-03-11
Adding optional expected type to unit validator so all units can be validated against

### LINKED ISSUE #663: 'Deg' or 'Degrees', or 'K' do not work in the units field edits
I think we are using the validator in length-units only mode, so we should change this to allow for other units too.

rerpha:
### Issue

Closes #663 

### Description of work

Adds an optional parameter when setting up a unit validator so an expected type can be specified. 
In the case of the add component window and for the OFF file units line edit, this is set to units of length (cm, m etc.)
Everywhere else this is now unset so stuff like `deg` will work. 

### Acceptance Criteria 

Added test for degrees etc without specifying an expected type. 

### UI tests

No ui tests added as problem was in logic anyway

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -99,7 +99,7 @@ def test_name_validator_set_to_duplicate_name():
     )
 
 
-def test_unit_validator_with_metres_set_as_expected_type():
+def test_unit_validator_with_metres_set_as_expected_dimensionality():
     validator = UnitValidator(expected_dimensionality=METRES)
 
     lengths = ["mile", "cm", "centimetre", "yard", "km"]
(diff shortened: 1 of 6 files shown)


tests/test_validators.py
```diff
@@ -126,6 +127,20 @@ def test_unit_validator():
         assert validator.validate(unit, 0) == QValidator.Intermediate
 
 
+def test_unit_validator_with_no_expected_type():
```
matthew-d-jones:
```suggestion
def test_unit_validator_with_no_expected_dimensionality():
```

tests/test_validators.py
```diff
@@ -98,8 +99,8 @@ def test_name_validator_set_to_duplicate_name():
     )
 
 
-def test_unit_validator():
-    validator = UnitValidator()
+def test_unit_validator_with_metres_set_as_expected_type():
```
matthew-d-jones:
```suggestion
def test_unit_validator_with_metres_set_as_expected_dimensionality():
```

PR #681, opened 2020-03-12
654 - Update filewriter control window to use new status message format

### LINKED ISSUE #654: Status message handling needs to be updated in the filewriter control window
The pane on the left is not working correctly so we should update it to use the new status message format.

### LINKED ISSUE #701: File-writer control window has no title
It is instead just called `nexus-constructor.py`. We should rename it to File-writer control window instead.

rerpha:
### Issue

Closes #654 
Closes #701 

### Description of work

Currently not fully able to be tested because the filewriter does not specify a service ID for each status message but everything else should still work.

### Acceptance Criteria 

- Refactored some duplicated code in `file_writer_control_window.py`
- Made Kafka consumer construct `File` and `Filewriter` objects as early as possible so that we don't need to try and parse JSON in multiple places 
- Refactored out helper functions where possible for easier testing

### UI tests

Adding a real model to existing tests so helpers can be tested. 


rerpha:
TODO: 
- Add filewriter even if it has no service ID, and test this 
- don't add a file if it has no file name - currently an empty status message will still get added to the list - test this.

matthew-d-jones:
This should wait until model_change is merged, but this branch needs to be updated to use the flatbuffer encoded status messages by making use of the ess-streaming-data-types library.
I also note that the `File` and `FileWriter` data classes being in kafka_interface.py looks odd.

PR #695, opened 2020-03-17
Model change part 1: Add component window

rerpha:
### Issue

None

### Description of work

Making a start on creating the model of the NC not using HDF. 
I have done this alongside the existing model so we can use it as reference and not have everything completely break. 

Have added some tests for the model helpers. 

Currently adding a component should work, however it will not show up in the tree view. It should be added to the component model. 
Fields should also work however attributes on those fields do not currrently work.

- [x] Nominate for code review


nexus_constructor/model/stream.py
```diff
@@ -0,0 +1,49 @@
+from abc import ABC
+
+import attr
+
+from nexus_constructor.writer_modules import WriterModules
+
+
+@attr.s
+class Stream(ABC):
```
rerpha: remove ABC - no methods on class so it does nothing

nexus_constructor/model/group.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC
+import attr
+
+from nexus_constructor.common_attrs import CommonAttrs
+from nexus_constructor.model.helpers import get_item, set_item
+
+
+@attr.s
+class Group(ABC):
```
rerpha: Remove ABC

nexus_constructor/model/component.py
```diff
@@ -0,0 +1,19 @@
+import attr
+
+from nexus_constructor.model.group import Group
+
+
+@attr.s
+class Component(Group):
+    @property
+    def description(self):
+        return self.children["description"] if "description" in self.children else None
```
rerpha: - use gettter/setter helper here
- use commonattrs here

nexus_constructor/model/helpers.py
```diff
@@ -0,0 +1,11 @@
+from typing import Dict, Any
+
+
+def get_item(dictionary_to_look_in: Dict[str, Any], attr_name: str) -> Any:
```
rerpha: fix + test this

rerpha:
I think this is now ready for review. I have added an attributes class so we can mirror the filewriter nexus structure in the model, which works quite nicely as it shares the "name" field like components and transformation which means `get_item()` and `set_item()` can be used.

matthew-d-jones:
@SkyToGround Is this along the lines of what you were thinking/expecting?

tests/test_model_helpers.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_get_item_works_as_expected_when_component_is_in_list():
+    name = "component1"
+    test = Component(name)
+
+    list_of_components = [test]
+    assert get_item(list_of_components, name) == test
+
+
+def test_get_item_works_as_expected_when_component_is_not_in_list_then_works_when_item_is_set():
```
matthew-d-jones: "works_as_expected" -> "throws_error"

nexus_constructor/model/transformation.py
```diff
@@ -0,0 +1,72 @@
...
+from PySide2.QtGui import QVector3D, QMatrix4x4
+import numpy as np
+
+from nexus_constructor.model.group import Group
+from nexus_constructor.model.helpers import get_item, set_item
+from nexus_constructor.common_attrs import CommonAttrs
+from nexus_constructor.transformation_types import TransformationType
+
+
+class TransformationGroup(Group):
```
SkyToGround: Is this class needed? I honestly do not remember enough of the NeXus documentation to be able to say.

nexus_constructor/model/transformation.py
```diff
@@ -0,0 +1,72 @@
...
+            QVector3D(vector[0], vector[1], vector[2]) if vector is not None else None
+        )
+
+    @vector.setter
+    def vector(self, new_vector: QVector3D):
+        vector_as_np_array = np.array([new_vector.x(), new_vector.y(), new_vector.z()])
+        set_item(self.attributes, CommonAttrs.VECTOR, vector_as_np_array)
+
+    @property
+    def ui_value(self):
```
SkyToGround: Is there a good reason as to why it is called `ui_value`? It is named simply `value` in the NeXus file, right? SO maybe use the same name here.

nexus_constructor/model/transformation.py
```diff
@@ -0,0 +1,72 @@
...
+        # TODO: get proper ui value
+        get_item(self.attributes, CommonAttrs.UI_VALUE)
+        return None
+
+    @property
+    def qmatrix(self) -> QMatrix4x4:
```
SkyToGround: It could be argued that 3D rendering code should be split from modelling code but it could also be argued that we should solve only one problem at a time. So we should probably leave this as is.

nexus_constructor/model/stream.py
```diff
@@ -0,0 +1,47 @@
+import attr
+
+from nexus_constructor.writer_modules import WriterModules
```
SkyToGround: I really should not point out that this file feels a bit incomplete as it is very obvious. 😃

nexus_constructor/model/helpers.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, List
+
+
+def find_item_index(list_to_look_in: List[Any], item_name: str):
```
SkyToGround: Do we need these helpers if we do not use HDF5 for data modelling?

nexus_constructor/model/stream.py
```diff
@@ -0,0 +1,47 @@
...
+
+HS00TYPES = ["uint32", "uint64", "float", "double"]
+
+
+class HS00Stream(Stream):
+    writer_module = WriterModules.HS00.value
+    data_type = attr.ib(type=str, validator=attr.validators.in_(HS00TYPES))
+    error_type = attr.ib(type=str, validator=attr.validators.in_(HS00TYPES))
+    edge_type = attr.ib(type=str, validator=attr.validators.in_(HS00TYPES))
+    shape = attr.ib(default=[])  # TODO: construct these objects
```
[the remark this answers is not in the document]
SkyToGround: Yes I agree with that one.

nexus_constructor/model/transformation.py
rerpha: `ui_value` is used for tracking the placeholder value for when the dataset does not have a value castable to float/int, ie a transformation having a stream group as its magnitude

nexus_constructor/model/component.py
```diff
@@ -0,0 +1,20 @@
...
+
+from nexus_constructor.model.group import Group
+from nexus_constructor.model.helpers import get_item, set_item
+
+
+@attr.s
+class Component(Group):
+    @property
+    def description(self):
+        return get_item(self.children, "description")
```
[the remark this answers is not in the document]
rerpha: Would you ever need to add a string to `children`? I would think that the only thing permitted in `children` is other objects. A string would be allowed in `values` but surely everything in children needs to have a name? besides stream groups?

`children` in the nexus structure is always a list, which means the only way to search through it is to use the `name` attribute. Is there a design issue here? What happens if you have two streams in a `children` list? how would they end up in the file?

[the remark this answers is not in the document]
matthew-d-jones: > Would you ever need to add a string to children?

No, I think my point was that you shouldn't be, but you are. If you make the change I suggest then I think a test will fall over and expose the problem.

nexus_constructor/model/transformation.py
SkyToGround: I am not convinced by that argument. On the other hand, it is easy to change the name latter so we can leave it as is for now.

rerpha:
I am going to try and get the "add component" window working, minus its editing an existing component functionality working in this PR, then call it day there to avoid having a humongous load of changes in one PR. 

currently i am nearly there, I just need to get the component shape working.

tests/model/test_model_component.py
```diff
@@ -0,0 +1,60 @@
...
+def test_component_set_description_correctly_sets_description():
+    comp = Component("comp3")
+    description = "a component"
+    comp.description = description
+
+    assert comp.description == description
+
+
+def test_component_set_field_with_numpy_array_correctly_sets_field_value():
+    import numpy as np
```
rerpha, line 30: Move this to top of class, no reason it should be in the test

rerpha:
I am calling it a day for this PR as it's quickly getting to the point where there are a lot of code changes but nothing is really working. I think shape information should be kept for another PR as its a bit of a can of worms in the scope of this. Currently nothing shows up in the component list, not sure why, but adding a component does not throw exceptions so i think the component model is being updated.

nexus_constructor/main_window.py
```diff
@@ -54,16 +54,16 @@ def setupUi(self, main_window):
...
-        self.treemodel = self.widget.findHdf5TreeModel()
-        self.treemodel.setDatasetDragEnabled(True)
-        self.treemodel.setFileDropEnabled(True)
-        self.treemodel.setFileMoveEnabled(True)
-        self.treemodel.insertH5pyObject(self.instrument.nexus.nexus_file)
+        # self.treemodel = self.widget.findHdf5TreeModel()
+        # self.treemodel.setDatasetDragEnabled(True)
+        # self.treemodel.setFileDropEnabled(True)
+        # self.treemodel.setFileMoveEnabled(True)
+        # self.treemodel.insertH5pyObject(self.instrument.nexus.nexus_file)
```
matthew-d-jones, line 61: Is there a ticket for sorting this out?

nexus_constructor/model/attribute.py
```diff
@@ -0,0 +1,16 @@
...
+import attr
+
+
+@attr.s
+class FieldAttribute:
+    """
+    Class for containing attributes in the nexus structure.
+    Uses the long format (dict with "name" and "values" as keys)
+    rather than the short form (name:value) as this works for everything and is more flexible, with
+    the trade-off being a longer message (which is not a priority)
```
matthew-d-jones, line 12: Need to mention the file writer JSON format to make this comment less confusing.

nexus_constructor/model/stream.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@attr.s
+class EV42Stream:
+    topic = attr.ib(type=str)
+    source = attr.ib(type=str)
+    _writer_module = attr.ib(type=str, default=WriterModules.EV42.value, init=False)
+
+
+# TODO: optional fields
```
matthew-d-jones, line 55: ticket not TODO

tests/model/test_model_helpers.py
```diff
@@ -0,0 +1,94 @@
+from nexus_constructor.model.component import Component
+from nexus_constructor.model.node import _get_item, _remove_item, _set_item
+import pytest
+
```
matthew-d-jones, line 4: test filename now doesn't match implementation filename

Issue #730, opened 2020-04-30
Get transformations working with new model change


rerpha:
Sub-ticket of #607 
Follows on from #695 

The hardest part of this is going to be the depends on chain as there isn't any real hierarchy now or at least not one that we can traverse easily because each transform doesn't currently contain a reference to its parent. 

The model has been mostly implemented in `model/transformations.py` but will need to be modified to contain the `depends_on` chain logic

Also implement a suitable replacement for checking equality between two transforms.

PR #738, opened 2020-05-15
730: transformations model change

DolicaAkelloEgwel:
### Issue

Makes progress on #730

### Description of work

- Replacing previous `has_link` behaviour in `TransformationsList`
- Using new `Transformation` in Component Tree Model and Component Tree View
- Implemented `transforms` and `transforms_full_chain` in Component
- Replacing `TransformationGroup` with `TransformationList`
- Implemented `depends_on` behaviour in Transformations
- Fixed some failing tests in test_transformation.py

### Acceptance Criteria 

Add a Translation or Rotation to a Component.
Check that the non-skipped Transformation tests pass.

### UI tests

n/a


nexus_constructor/model/transformation.py
```diff
@@ -9,13 +11,14 @@
 from nexus_constructor.transformation_types import TransformationType
 
 
+@attr.s
 class TransformationGroup(Group):
```
rerpha, line 15: This can be removed now can't it?

nexus_constructor/model/transformation.py
```diff
@@ -46,9 +51,21 @@ def vector(self, new_vector: QVector3D):
 
     @property
     def ui_value(self) -> float:
-        if isinstance(self.values, (float, int)):
-            return float(self.values)
-        return float(self.get_attribute_value(CommonAttrs.UI_VALUE))
+
+        ui_value = self.get_attribute_value(CommonAttrs.UI_VALUE)
```
rerpha, line 55: I think this should be checking the magnitude of the dataset value first, rather than the UI value attribute, and seeing if it is an int/double. Then if it's anything else it should revert to using the `ui_value` attribute.

nexus_constructor/transformation_view.py
```diff
@@ -62,6 +63,7 @@ def enable(self):
     def saveChanges(self):
         self.transformation.ui_value = self.transformation_frame.value_spinbox.value()
         # self.transformation.dataset = self.transformation_frame.magnitude_widget.value
```
rerpha, line 65: ah ok i now see why the code in ui_value is different. ignore my previous comment about incorrect logic, but could you create an issue to get transformation magnitudes and `ui_value`  working properly please?

PR #742, opened 2020-05-20
730: Preserving distance information

DolicaAkelloEgwel:
### Issue

Progresses #730

### Description of work

Writing/loading distance information on Transformations should now be 100% working for Scalar, Array, and Stream data. Advanced stream info should also work as well.

### Acceptance Criteria 

Write distance information of these types and check that exiting and entering edit mode for the Transformation doesn't cause the input to disappear.

### UI tests

n/a


nexus_constructor/model/stream.py
```diff
@@ -50,6 +50,11 @@ class EV42Stream:
     topic = attr.ib(type=str)
     source = attr.ib(type=str)
     writer_module = attr.ib(type=str, default=WriterModules.EV42.value, init=False)
+    adc_pulse_debug = attr.ib(type=bool, default=None)
```
rerpha, line 53: I think these may need to match the names of the ones in the filewriter JSON, jonas said he wanted to change them to a flat structure so `nexus.indices.index_every_mb` is valid. perhaps we should just use underscores in place of the `.`s

also, i think some of them are booleans rather than ints, so the type should match in the `attr.ibs` declaration. the documentation [here](https://github.com/ess-dmsc/kafka-to-nexus/tree/v3.0.0/documentation) should hopefully help, i can't remember which ones were booleans off the top of my head

nexus_constructor/model/transformation.py
```diff
@@ -84,6 +84,14 @@ def units(self):
     def units(self, new_units):
         self.set_attribute_value(CommonAttrs.UNITS, new_units)
 
+    @property
+    def value(self) -> Dataset:
```
rerpha, line 88: i think instead of using a property here we should use the `values` member of the dataset, using an extra attribute doesn't match the JSON. unless you think it is easier to use a property?

nexus_constructor/stream_fields_widget.py
```diff
@@ -335,44 +336,39 @@ def get_stream_group(self) -> StreamGroup:
...
-            self._create_dataset_from_spinner(
-                stream_group, self.ev42_nexus_to_spinner_ui_element
-            )
-
-    def _create_f142_fields(self, stream_group: F142Stream):
+        stream.index_every_mb = self.f142_nexus_to_spinner_ui_element[
+            NEXUS_INDICES_INDEX_EVERY_MB
+        ].value()
+        stream.index_every_kb = self.f142_nexus_to_spinner_ui_element[
+            NEXUS_INDICES_INDEX_EVERY_KB
```
rerpha, line 348: I wonder if we should just scrap the `f142_nexus_to_spinner_ui_element` dict and use the value from the UI element in `self` here. it doesn't really make things look neat with the new model change and would be much clearer to just call something like this: 
`stream.index_every_mb = self.index_every_mb_spinbox.value()`

nexus_constructor/stream_fields_widget.py
```diff
@@ -335,44 +336,39 @@ def get_stream_group(self) -> StreamGroup:
...
-            )
-        if self.advanced_options_enabled:
-            self._create_dataset_from_spinner(
-                stream_group, self.f142_nexus_to_spinner_ui_element
-            )
+        stream.adc_pulse_debug = self.ev42_adc_pulse_debug_checkbox.isChecked()
+        stream.index_every_mb = self.ev42_nexus_to_spinner_ui_element[
+            NEXUS_INDICES_INDEX_EVERY_MB
+        ].value()
+        stream.index_every_kb = self.ev42_nexus_to_spinner_ui_element[
```
rerpha, line 363: as above

nexus_constructor/stream_fields_widget.py
```diff
@@ -422,17 +435,31 @@ def fill_in_existing_f142_fields(self, field: F142Stream):
...
+            self._fill_existing_advanced_f142_fields(field)
+
+    def _fill_existing_advanced_f142_fields(self, field: F142Stream):
+        """
+        Fill the advanced fields in the interface with the existing f142 stream data.
+        :param field: The f412 stream data object.
+        """
+        self.f142_nexus_to_spinner_ui_element[NEXUS_INDICES_INDEX_EVERY_MB].setValue(
+            field.index_every_mb
+        )
```
rerpha, line 451: same as above but calling setValue() instead

nexus_constructor/transformation_view.py
```diff
@@ -30,19 +30,18 @@ def _fill_in_existing_fields(self, current_vector):
...
-        #     )
-        # if isinstance(self.transformation, NXLogTransformation):
-        #     self.transformation_frame.main_layout.addWidget(
-        #         QLabel(
-        #             "Transformation is an NXlog - currently these are not editable. "
-        #         )
-        #     )
-        #     self.transformation_frame.magnitude_widget.setVisible(False)
-        # Disabled whilst working on model change
+        update_function = find_field_type(self.transformation.value)
```
rerpha, line 33: could call `self.transformation.values` here instead and remove the `value` property, as  `Transformation` inherits from `Dataset` anyway.

nexus_constructor/model/transformation.py
rerpha: I think we should copy the previous `UI_value` functionality as well in the old transformations class (ui value is dataset if scalar, float/int castable, else use the ui value stored as an attribute)

nexus_constructor/stream_fields_widget.py
```diff
@@ -388,28 +384,45 @@ def fill_in_existing_ev42_fields(self, field: EV42Stream):
...
-            )
+    def _fill_existing_advanced_ev42_fields(self, field: EV42Stream):
+        """
+        Fill the fields in the interface with the existing ev42 stream data.
+        :param field: The ev42 stream data object.
+        """
+        self.ev42_adc_pulse_debug_checkbox.setChecked(field.adc_pulse_debug)
+        self.ev42_nexus_to_spinner_ui_element[NEXUS_INDICES_INDEX_EVERY_MB].setValue(
+            field.index_every_mb
+        )
```
rerpha, line 411: Same as above here too.

nexus_constructor/transformation_view.py
```diff
@@ -30,19 +30,18 @@ def _fill_in_existing_fields(self, current_vector):
...
+        if update_function is not None:
+            update_function(
+                self.transformation.value, self.transformation_frame.magnitude_widget
+            )
+        if isinstance(self.transformation, NXLogTransformation):
```
rerpha, line 38: lets just remove this conditional, nxlogs are not going to work at the moment as we can't load in from NeXus files and there's no way to create them in the UI.

nexus_constructor/stream_fields_widget.py
matthew-d-jones: I agree. Alternatively, you could stay closer to how this was implemented before by using setattr:
```python
setattr(stream, NEXUS_INDICES_INDEX_EVERY_MB.split('.')[-1], self.f142_nexus_to_spinner_ui_element[NEXUS_INDICES_INDEX_EVERY_MB].value())
```
which would allow a more similar approach to before with looping through the dictionary.

PR #743, opened 2020-05-21
Model change cleanup

rerpha:
### Issue

None

### Description of work

Trying to get the jenkins build to pass. currently failing on flake8 so have tidied the code up and removed any stuff we don't need anymore. 

### Acceptance Criteria 

Things that should work by now should still work. there shouldn't really be any functional changes here  just removing stuff that wasn't used by anything other than tests anyway. 

### UI tests

Some tests for the old model have been removed in accordance to the model components being removed. 


nexus_constructor/component/component_type.py
```diff
@@ -1,5 +1,6 @@
-from typing import List
 import os
+from typing import List
+
```
matthew-d-jones, line 3: whitespace?

nexus_constructor/component/chopper_shape.py
```diff
@@ -1,17 +1,18 @@
 import logging
+from typing import Optional, Union, List, Tuple
 
+from PySide2.QtGui import QVector3D
 from nexus_constructor.component.component_shape import ComponentShape
 from nexus_constructor.geometry.cylindrical_geometry import CylindricalGeometry
+
```
matthew-d-jones, line 7: whitespace?

.flake8
```diff
@@ -7,3 +7,4 @@ exclude = ui/,build*,definitions
 per-file-ignores =
     nexus_constructor/geometry/__init__.py:F401
     tests/ui_tests/test_ui_fields.py:F811,F401
+    tests/*:E402,F821
```
matthew-d-jones, line 10: What are these for? Are you allowed comments in this config file? Would be nice to record why these codes are ignored.

rerpha: Have added a comment to explain in the flake8 file, did try it previously but the comment caused issues. seems like you have to put the comment on a separate line in the .flake8 file

PR #747, opened 2020-05-28
Make default translation along z

### LINKED ISSUE #715: Make the added transformation when clicking "add transformation" increment along the z-axis
Currently when a new transformation is added it gets placed at 0,1,0 from the component's position along the x-axis. Instead it should be put at 0,0,1 instead.

matthew-d-jones:
### Issue

Closes #715

### Description of work

Make default transformation dtype "Double" and default translation vector along z axis.

### Acceptance Criteria 

Make a new translation and check value dtype is "Double" and vector is (0, 0, 1).

### DIFF
--- a/nexus_constructor/component_tree_model.py
+++ b/nexus_constructor/component_tree_model.py
@@ -225,14 +225,14 @@ class ComponentTreeModel(QAbstractItemModel):
         parent_component, transformation_list, transformation_type
     ):
         values = Dataset(
-            name="", dataset=DatasetMetadata(type="Byte", size=[1]), values=""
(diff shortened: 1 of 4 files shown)


nexus_constructor/component_tree_model.py
```diff
@@ -225,14 +225,14 @@ def _create_new_transformation(
         parent_component, transformation_list, transformation_type
     ):
         values = Dataset(
-            name="", dataset=DatasetMetadata(type="Byte", size=[1]), values=""
+            name="", dataset=DatasetMetadata(type="Double", size=[1]), values=""
```
rerpha, line 228: We should change the `values=""` to `values=1` I think.

PR #752, opened 2020-05-28
#745: Fixing external transformations

### LINKED ISSUE #730: Get transformations working with new model change

### LINKED ISSUE #740: Rename Component.transform property
Rename to something more helpful, perhaps qtransform or final_3dview_transform.

Add return type hints for `transform` and `transforms`.

### LINKED ISSUE #745: Model change: Fix external transformation dependencies
The transform depends_on chain does not fully work yet as external transformations (ie having a depends_on as another transformation in a separate component) do not work.

rerpha:
### Issue

Closes #745 
Closes #740 
Closes #730 

### Description of work

- Fixes `depends_on` chain for local transformations 
- Allows external transformations by using a reference to the depended on component in `LinkTransformation`
- Removes horrible sample UI logic and treats that as a component, to simplify the 3d view. 

### Acceptance Criteria 

- Adding a link should work: add component1, add transformation, add component2, add link to component1, add transformation, should be dependent on the first component's position.
- Deleting a link should also work - follow steps above but delete link, second component should be moved to position relative to origin/sample rather than 

### UI tests

- Updated to use new model
- Some unit tests enabled and fixed


rerpha:
still wip, currently all transforms are broken while i try and sort out the depends_on chain for everything including links and local transforms (local transforms are broken i think)

nexus_constructor/component_tree_model.py
```diff
@@ -71,15 +71,15 @@ def _get_transformation_list(self, node, parent_item):
         transformation_list = None
         target_index = QModelIndex()
         if isinstance(parent_item, Component):
-            if not hasattr(parent_item, "stored_transforms"):
-                parent_item.stored_transforms = parent_item.transforms
-            transformation_list = parent_item.stored_transforms
+            # if not hasattr(parent_item, "stored_transforms"):
+            #     parent_item.stored_transforms = parent_item.transforms
```
matthew-d-jones, line 75: remove?

nexus_constructor/instrument_view.py
```diff
@@ -200,7 +200,10 @@ def add_component(
 
         mesh = OffMesh(geometry.off_geometry, self.component_root_entity, positions)
         material = create_material(
-            QColor("black"), QColor("grey"), self.component_root_entity
+            QColor("black") if name != "sample" else QColor("red"),
```
matthew-d-jones, line 203: Use `SAMPLE_NAME` from model.instrument

nexus_constructor/model/attribute.py
```diff
@@ -13,7 +14,13 @@ class FieldAttribute:
     """
 
     name = attr.ib(type=str)
-    values = attr.ib(type=ValueType)
+    values = attr.ib(type=ValueType, cmp=False)
+
+    def __eq__(self, other_attribute):
```
matthew-d-jones, line 19: Would be good to cover with unit test

PR #756, opened 2020-06-01
736 pixel data

### LINKED ISSUE #736: Fix pixel data in new model change
Follows on from #735 

Pixel data is not currently working, it is imported but is not yet used when setting the shape properties in `model/component.py`

DolicaAkelloEgwel:
### Issue

Closes #736

### Description of work

Restores pixel data behaviour.

### Acceptance Criteria 

Check that creating and modifying pixel data works. Check that the pixel options tests pass.

### UI tests

n/a

### DIFF
--- a/nexus_constructor/model/component.py
+++ b/nexus_constructor/model/component.py
@@ -222,7 +222,8 @@ class Component(Group):
     ):
         self.remove_shape()
 
-        geometry = OFFGeometryNexus(self._get_shape_group_for_pixel_data(pixel_data))
+        shape_group = self._get_shape_group_for_pixel_data(pixel_data)
+        geometry = OFFGeometryNexus(shape_group)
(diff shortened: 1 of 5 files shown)


nexus_constructor/model/component.py
```diff
@@ -303,6 +303,32 @@ def record_pixel_mapping(self, pixel_mapping: PixelMapping):
             "int64",
         )
 
+    def _create_transformation_vectors_for_pixel_offsets(self) -> List[QVector3D]:
+        """
+        Construct a transformation (as a QVector3D) for each pixel offset
+        """
+        x_offsets = self.get_field_value("x_pixel_offset")
```
rerpha, line 310: could this throw? in which case i think wrap all three in a ValueError try/except

DolicaAkelloEgwel: I've caught the exception in `get_field_value`

PR #759, opened 2020-06-02
Get field value fix

DolicaAkelloEgwel:
### Issue

Sorts out the field value issue.

### Description of work

Restored the original `get_field_value` behaviour and created a test for it. Also extracted `_get_shape_group_for_pixel_data` and gave it a doc string.

### Acceptance Criteria 

Check that the tests pass.

### UI tests

n/a


tests/test_group.py
```diff
@@ -0,0 +1,11 @@
+import pytest
```
rerpha, line 1: can this file be moved to `tests/model/` please?

nexus_constructor/model/component.py
```diff
@@ -307,14 +318,16 @@ def _create_transformation_vectors_for_pixel_offsets(self) -> List[QVector3D]:
...
         """
-        x_offsets = self.get_field_value("x_pixel_offset")
-        y_offsets = self.get_field_value("y_pixel_offset")
-        z_offsets = self.get_field_value("z_pixel_offset")
-        if x_offsets is None or y_offsets is None:
+        try:
+            x_offsets = self.get_field_value("x_pixel_offset")
+            y_offsets = self.get_field_value("y_pixel_offset")
+        except AttributeError:
             raise Exception(
```
rerpha, line 325: Should we raise an exception here? or can we just log the error and `return`?

nexus_constructor/model/component.py
```diff
@@ -46,6 +46,17 @@ def _normalise(input_vector: QVector3D) -> Tuple[QVector3D, float]:
     return input_vector.normalized(), magnitude
 
 
+def _get_shape_group_for_pixel_data(pixel_data: Type[PixelData]) -> str:
```
matthew-d-jones, line 49: The type hint here should be `PixelData` not `Type[PixelData]`. `Type` is used when you pass a class object rather than an instance of a class.

nexus_constructor/model/component.py
matthew-d-jones: Raising an exception looks right to me. The method is only used where you expect and need a result and there isn't a sensible course of action if it can't return something.

PR #766, opened 2020-06-09
Try using 7zip for archiving on windows

matthew-d-jones:
~Merge #765 first.~

Archiving on Windows was very slow ~5 mins. I've installed a 7zip powershell cmdlet and using it reduces the archiving to ~2 mins.


rerpha:
#765 merged

PR #788, opened 2020-06-17
782: Load from JSON prototype

### LINKED ISSUE #782: Load from JSON prototype
Create prototype capable of loading a couple of model classes.
Could try using `object_hook` argument to directly create the model with call to `json.load()`: https://docs.python.org/3/library/json.html#json.load
Or load to dictionaries and then traverse the nested dictionaries creating model objects. (each model object could optionally take a dictionary to initialise itself with?)

DolicaAkelloEgwel:
### Issue

Closes #782

### Description of work

Allows basic reading of JSON files. Currently only capable of discovering the names and nxclasses of components.

### Acceptance Criteria 

Check that the tests pass. Create a file with a sample and component and check that loading the file recovers the sample name and component name + class.

### UI tests

n/a

### DIFF
--- a/nexus_constructor/model/load_from_json.py
+++ b/nexus_constructor/model/load_from_json.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from typing import Union
 
 from PySide2.QtWidgets import QWidget
 
@@ -12,14 +13,20 @@ from nexus_constructor.ui_utils import show_warning_dialog
 IGNORE = ["entry", "transformations", "NX_class"]
 
 
-def _read_nx_class(entry: list) -> str:
-    for item in entry:
-        if item.get("name") == "NX_class":
-            return item.get("values")
-        if item.get("NX_class"):
-            return item.get("NX_class")
-
+def _find_nx_class(entry: dict) -> str:
+    if entry.get("name") == "NX_class":
+        return entry.get("values")
+    if entry.get("NX_class"):
+        return entry.get("NX_class")
     return ""
+
+
+def _read_nx_class(entry: Union[list, dict]) -> str:
+    if isinstance(entry, list):
+        for item in entry:
+            return _find_nx_class(item)
+    elif isinstance(entry, dict):
+        return _find_nx_class(entry)
 
 
 def _read_transformations(entry: list):
@@ -86,15 +93,15 @@ class JSONReader:
 
         name = json_object.get("name")
 
-        if name == "instrument":
-            return True
-
-        elif name and name not in IGNORE:
+        if name:
 
             nx_class = _read_nx_class(json_object.get("attributes"))
 
             if not _validate_nx_class(nx_class):
                 return False
+
+            if name == "NXinstrument":
+                return True
 
             if nx_class == "NXsample":
                 component = self.entry.instrument.sample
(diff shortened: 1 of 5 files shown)


nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,118 @@
...
+    for item in entry:
+        if item.get("name") == "transformations":
+            return item.get("children")
+
+    return None
+
+
+def _retrieve_children_list(json_dict: dict) -> list:
+    try:
+        return json_dict["nexus_structure"]["children"][0]["children"]
```
matthew-d-jones, line 35: I think this makes it a little clearer:
```suggestion
        entry = json_dict["nexus_structure"]["children"][0]
        return entry["children"]
```

nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,118 @@
...
+                )
+                return False
+
+            children_list = _retrieve_children_list(json_dict)
+
+            if not children_list:
+                logging.info("Unable to locate components in JSON file.")
+                return False
+
+            return all(self._read_json_object(child) for child in children_list)
```
matthew-d-jones, line 83: Rather than failing if we come across anything we don't know what to do with in the file, we should load what we can and accumulate warnings that can be displayed to the user in one go at the end of the parsing process.

nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _read_json_object(self, json_object: dict) -> bool:
+
+        name = json_object.get("name")
+
+        if name == "instrument":
```
matthew-d-jones, line 89: In the NeXus standard groups can be named anything, test for the "NX_class" attribute being "NXinstrument" instead.

nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,118 @@
...
+            return item.get("values")
+        if item.get("NX_class"):
+            return item.get("NX_class")
+
+    return ""
+
+
+def _read_transformations(entry: list):
+    for item in entry:
+        if item.get("name") == "transformations":
```
matthew-d-jones, line 27: I think this is good enough for the initial implementation, but just so that you are aware:
This will work for JSON output from the NeXus Constructor but there is no reason in the NeXus format that this has to be called "transformations".

nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,155 @@
...
+                return False
+
+            for child in children_list:
+                self._read_json_object(child)
+
+            if self.warnings:
+                show_warning_dialog(
+                    "\n".join(self.warnings), "Invalid JSON", parent=self.parent,
+                )
+                return False
```
matthew-d-jones, line 102: We should still display what we have managed to load
```suggestion
                return True
```

nexus_constructor/model/load_from_json.py
```diff
@@ -0,0 +1,155 @@
...
+                    parent=self.parent,
+                )
+                return False
+
+            for child in children_list:
+                self._read_json_object(child)
+
+            if self.warnings:
+                show_warning_dialog(
+                    "\n".join(self.warnings), "Invalid JSON", parent=self.parent,
```
matthew-d-jones, line 100: The JSON itself isn't necessarily invalid, it's just that some parts of it don't match what we expect the File Writer to parse successfully.
```suggestion
                    "\n".join(self.warnings), "Warnings encountered loading JSON", parent=self.parent,
```

Issue #757, opened 2020-06-02
Fix Loading in from JSON to model


rerpha:
~Prototype, ticket #782 first!~

We need to be able to load in from a JSON file to the model. We should test this with one that we've saved as well as one of the [filewriter system test commands](https://github.com/ess-dmsc/kafka-to-nexus/tree/master/system-tests/commands)

Remaining subtasks:
- #798
- #799 
- #805

PR #792, opened 2020-06-23
757: Some work on reading Transformations

DolicaAkelloEgwel:
### Issue

Progresses #757 

### Description of work

Added ability to read Transformations from a JSON file and add this to the Component `transforms_list` attribute. Doesn't handle `depends_on` behaviour yet. For the time being I've set the component depends on to one of its Transformations. Works for the Instrument View but not the Component Tree View.

### Acceptance Criteria 

Look through _create_transformations and check that you agree with what its doing. Also check that the tests pass and that the new tests cover appropriate test cases.

### UI tests

n/a


nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+    :param name: The transformation name.
+    :return: A dataset containing the above information.
+    """
+    return Dataset(
+        name, dataset=DatasetMetadata(size=[1], type=dtype), values=angle_or_magnitude,
+    )
+
+
+class TransformationReader:
+    def __init__(self, parent_component: Component, entry: list):
```
matthew-d-jones, line 97: This would benefit from a doc comment. "entry" here is "children of the component entry".

nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+                    f" {transform_name}"
+                )
+                f" {self.parent_component.name}."
+            else:
+                msg = f"Cannot find {property_name} for transformation in component"
+                f" {self.parent_component.name}."
+            self.warnings.append(msg)
+            return failure_value
+
+    def _find_property_in_list(
```
matthew-d-jones, line 150: If you pass the warning message string into the method then couldn't this method also be used to find "NX_class" instead of having a separate implementation for that (`_read_nx_class`)?

nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+        self,
+        property_name: str,
+        transformation_name: str,
+        attributes_list: list,
+        failure_value: Any = None,
+        parse_result_func: Callable = None,
```
matthew-d-jones, line 156: I can't spot a benefit to passing `parse_result_func` into this method. Why not simply call it on the result?

nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+                except KeyError:
+                    continue
+
+    def _get_transformation_property(
+        self,
+        property_name: str,
+        json_transformation: dict,
+        transform_name: str = None,
+        failure_value: Any = None,
+        parse_result_func: Callable = None,
```
matthew-d-jones, line 120: I can't spot a benefit to passing `parse_result_func` into this method. Why not simply call it on the result?

nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+        """
+        for attribute in attributes_list:
+            try:
+                if attribute["name"] == property_name:
+                    if parse_result_func:
+                        return parse_result_func(attribute["values"])
+                    else:
+                        return attribute["values"]
+            except KeyError:
+                continue
```
matthew-d-jones, line 175: If I follow correctly, if you reach this line then you've found the property with the specified name but it had no "values" field. I don't think it makes sense to continue and try to look for another property with the same name:
`break` rather than `continue`?

nexus_constructor/model/load_from_json.py
matthew-d-jones: There is some inconsistency in terminology, does "property" always mean "attribute"?
If yes then you could rename the method, if no then `attributes_list` could be renamed.

nexus_constructor/model/load_from_json.py
```diff
@@ -53,9 +78,217 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+    :param dtype: The data type.
+    :param name: The transformation name.
+    :return: A dataset containing the above information.
+    """
+    return Dataset(
+        name, dataset=DatasetMetadata(size=[1], type=dtype), values=angle_or_magnitude,
+    )
+
+
+class TransformationReader:
```
matthew-d-jones, line 96: I think it would be beneficial to move this class to a new file.

PR #793, opened 2020-06-25
786 missing as dicts

### LINKED ISSUE #774: Array field with single row is converted to a column array after creating component
Editing a component and then editing the array shows that it has become transposed.

### LINKED ISSUE #786: Create as_dict implementation for things that are not yet being serialised to JSON
Things that need investigating and potentially doing:
- Shape
- Pixel data 
- Streams?
- Links
- Transformations
- Array attributes
- Check File Writer command created correctly (create separate ticket if more work involved)

And we should add tests for all of the above

rerpha:
### Issue

Closes #786 
Closes #774 

### Description of work

Adds `as_dict()` implementation in order to be able to serialise the model to JSON. 

- ~Shape~
- ~Pixel data~ 
- ~Streams~
- ~Links~
- ~Transformations~
- ~Array attributes~

### Acceptance Criteria 

Streams, transforms, links, arrays, scalar datasets are all present in the outputted JSON. 

### UI tests

Unit tests created to check JSON output is correct


nexus_constructor/model/group.py
```diff
@@ -54,6 +55,18 @@ def as_dict(self) -> Dict[str, Any]:
         return {
             "name": self.name,
             "type": self.type,
-            "attributes": [attribute.as_dict() for attribute in self.attributes],
-            "children": [child.as_dict() for child in self.children],
+            "attributes": [
+                attribute.as_dict()
+                for attribute in self.attributes
+                if attribute.name != CommonAttrs.DEPENDS_ON
```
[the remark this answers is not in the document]
rerpha: in that case i think we ought to change the `depends_on` setters and getters for components, as they are currently incorrect

rerpha:
nb: just reverted a commit that removed `_parent_component` from transformations. we still need this, as the current transformation's `parents` are actually the transformation group

rerpha:
To do: move ui_value from being an attribute and instead make it a member variable

matthew-d-jones:
If there are no attributes the field should be omitted or have an empty array, I think "null" will cause problems at the filewriter.
The missing value of description being an empty array is also odd and likely to be a problem. I think it should be an empty string or we omit the dataset if it is empty.
```
{
                  "name": "description",
                  "type": "dataset",
                  "attributes": null,
                  "values": []
                },
```

nexus_constructor/model/dataset.py
```diff
@@ -37,5 +37,5 @@ def as_dict(self) -> Dict[str, Any]:
             "attributes": [attribute.as_dict() for attribute in self.attributes]
             if self.attributes
             else None,
-            "values": values if values else [],
+            "values": values if values else "",
```
matthew-d-jones, line 40: I don't think this is the right solution. We'd want "" if it is a string, [] if array and so on. Perhaps this should just be
```
"values": values
```

nexus_constructor/model/group.py
```diff
@@ -82,6 +79,11 @@ def as_dict(self) -> Dict[str, Any]:
             if self.children
             else [],
         }
+        if self.attributes:
+            return_dict["attributes"] = [
+                attribute.as_dict() for attribute in self.attributes
+            ]
+        return return_dict
```
matthew-d-jones, line 86: I'm still seeing `"attributes": null,` in some places. My guess would be that the problem is that `self.attributes` can be true but the only attribute that is present is in the exclude list. Solution would be to make the list of attributes first and then test if it is empty.

PR #794, opened 2020-06-25
Fix freezing app on windows

### LINKED ISSUE #733: Remove nexus tree view
Sub-ticket of #607
Follows on from #695

This may require the ability to save to nexus beforehand.

### LINKED ISSUE #789: Windows distributable fails to run
Error due to `importlib` being missing from the frozen app.
I think `importlib` is a new dependency of flatbuffers and the problem was introduced when we updated the version of flatbuffers on the Windows Jenkins node.

matthew-d-jones:
### Issue

Closes #733 
Closes #789

### Description of work

Use pyinstaller rather then cx_freeze to make distributable packages of the NeXus Constructor.
Fixes Windows distributable. Also completes ticket to remove silx view as that was causing a problem with the distributable anyway.

Other nice side effects:
- On both Linux and Windows the package is more than 100MB smaller than was produced with cx_freeze.
- The Linux package created on CentOS runs successfully for me on Ubuntu.

### Acceptance Criteria 

Check distributable packages for Linux and Windows work.


matthew-d-jones:
installing dependencies in virtualenv on windows is taking >12 minutes, see if repeatable problem. Maybe machine is running updates or something?

Issue #799, opened 2020-06-30
JSON Loading: Depends on


DolicaAkelloEgwel:
Saving and/or loading of transformations with a `depends_on` defined is not working.

For example: create two components and create a translation in one component which depends on a translation in the other component. Save to JSON and then try to load the JSON, you will get an error something like this:
```
Traceback (most recent call last):
  File "/home/matt/git/nexus-constructor/nexus_constructor/main_window.py", line 161, in open_json_file
    success = reader.load_model_from_json(filename)
  File "/home/matt/git/nexus-constructor/nexus_constructor/json/load_from_json.py", line 281, in load_model_from_json
    self.component_dictionary[dependency_component_name],
KeyError: 'Translation'
```

PR #801, opened 2020-06-30
799: JSON component depends on

DolicaAkelloEgwel:
### Issue

Progresses #799

### Description of work

Program should now identify the transformation object that a component depends on in order to set its depends_on attribute. Haven't done anything for transformations depending on other transformations yet. 

### Acceptance Criteria 

*What you have changed, how the reviewer should test it*

### UI tests

n/a

### DIFF
--- a/nexus_constructor/json/load_from_json.py
+++ b/nexus_constructor/json/load_from_json.py
@@ -32,14 +32,6 @@ def _retrieve_children_list(json_dict: dict) -> list:
         return []
 
 
-def _get_transformation_by_name(
-    component: Component, transformation_name: str
-) -> Transformation:
-    for transformation in component.transforms_list:
-        if transformation.name == transformation_name:
-            return transformation
-
-
 class JSONReader:
     def __init__(self, parent: QWidget):
         self.entry = Entry()
@@ -48,6 +40,23 @@ class JSONReader:
         self.warnings = []
         self.depends_on_paths = dict()
         self.component_dictionary = dict()
+
+    def _get_transformation_by_name(
+        self, component: Component, transformation_name: str
+    ) -> Transformation:
+        """
+        Finds a transformation in a component based on its name.
+        :param component: The component the transformation belongs to.
+        :param transformation_name: The name of the transformation.
+        :return: The transformation with the given name.
+        """
+        for transformation in component.transforms_list:
+            if transformation.name == transformation_name:
+                return transformation
+        self.warnings.append(
+            f"Unable to find transformation with name {transformation_name} in component {component.name} in order to "
+            f"set depends_on value."
+        )
 
     def load_model_from_json(self, filename: str) -> bool:
         """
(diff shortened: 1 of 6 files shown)


nexus_constructor/json/load_from_json.py
```diff
@@ -28,12 +32,22 @@ def _retrieve_children_list(json_dict: dict) -> list:
...
+            return transformation
+
+
 class JSONReader:
     def __init__(self, parent: QWidget):
         self.entry = Entry()
         self.entry.instrument = Instrument()
         self.parent = parent
         self.warnings = []
+        self.depends_on_paths = dict()
```
matthew-d-jones, line 41: Type hint and comment to explain what goes in these dictionaries would be really useful to aid understanding.
Here is an example for `self.depends_on_paths`, please do the same for `self.component_dictionary`.
```suggestion
        # key: component name, value: NeXus path pointing to transformation that component depends on
        self.depends_on_paths: Dict[str, str] = dict()
```

nexus_constructor/json/load_from_json.py
```diff
@@ -3,14 +3,18 @@
...
+    _find_nx_class,
+    _find_attribute_from_list_or_dict,
+    DEPENDS_ON_IGNORE,
+)
 from nexus_constructor.json.transformation_reader import TransformationReader
 from nexus_constructor.model.component import Component
 from nexus_constructor.model.entry import Entry
 from nexus_constructor.model.instrument import Instrument
+from nexus_constructor.model.transformation import Transformation
 from nexus_constructor.ui_utils import show_warning_dialog
```
matthew-d-jones, line 17: I think it would be a good idea to use a doc comment to record some assumptions that are made in this implementation.
```suggestion

"""
The current implementation makes a couple of assumptions that may not hold true for all valid JSON descriptions of valid NeXus files, but are safe if the JSON was created by the NeXus Constructor:
1. All transformations exist in NXtransformations groups inside components.
2. All depends_on paths are absolute, not relative.
"""
```

nexus_constructor/json/load_from_json.py
```diff
@@ -71,6 +85,18 @@ def load_model_from_json(self, filename: str) -> bool:
                     child, json_dict["nexus_structure"]["children"][0].get("name")
                 )
 
+            for key in self.depends_on_paths.keys():
+                depends_on_path = self.depends_on_paths[key].split("/")[3:]
```
matthew-d-jones:
```suggestion
                # The following extraction of the component name and transformation name makes the assumption
                # that the transformation lives in a component and nowhere else in the file, this is safe assuming
                # the JSON was created by the NeXus Constructor.
                depends_on_path = self.depends_on_paths[key].split("/")[3:]
```

nexus_constructor/json/load_from_json.py
matthew-d-jones: @DolicaAkelloEgwel if you address this comment (document `self.component_dictionary`) then I think we could merge what you have done so far.

PR #815, opened 2020-07-23
814 Fix set_field_value by removing optional dtype and making dtype required

### LINKED ISSUE #814: Remove optional dtype argument for set_field_value`
Dtype should always be specified, so make it required in the interface and change the dtypes of anywhere it has not been set accordingly.

The nexus constructor is also currently not outputting type information as the as_dict is not called for `DatasetMetadata` so we should fix that too.

rerpha:
### Issue

Closes #814 

### Description of work

Makes `dtype` a required argument for `set_field_value()` and replaces any instances where it was used without the argument and provides one. 

### Acceptance Criteria 

- Everything still works and tests still pass.
- Dtypes that have been set are of the correct type

### UI tests

Some changed to provide the dtype arg if not already doing so. 

### DIFF
--- a/nexus_constructor/model/dataset.py
+++ b/nexus_constructor/model/dataset.py
@@ -4,7 +4,22 @@ import numpy as np
 
 from nexus_constructor.common_attrs import CommonAttrs
 from nexus_constructor.model.node import Node
-from nexus_constructor.model.value_type import ValueType
+from nexus_constructor.model.value_type import ValueType, VALUE_TYPE
+
+
+def _get_human_readable_type(new_value: Any):
+    if isinstance(new_value, list):
+        new_value = new_value[0]
+    if isinstance(new_value, str):
+        return "String"
+    elif isinstance(new_value, int):
+        return "Int"
+    elif isinstance(new_value, float):
+        return "Double"
+    else:
+        return next(
+            key for key, value in VALUE_TYPE.items() if value == new_value.dtype
+        )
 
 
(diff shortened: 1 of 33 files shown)


nexus_constructor/model/dataset.py
```diff
@@ -4,20 +4,52 @@
...
+        size=len(ds.values) if isinstance(ds.values, list) else "1",
+        type=_get_human_readable_type(ds.values),
+    )
+
 
 @attr.s
 class Dataset(Node):
-    dataset = attr.ib(type=DatasetMetadata, default={})
     values = attr.ib(factory=list, type=List[ValueType])
     type = attr.ib(type=str, default="dataset", init=False)
```
rerpha, line 49: remove this and just put it in as_dict

nexus_constructor/model/dataset.py
```diff
@@ -4,20 +4,52 @@
 
 from nexus_constructor.common_attrs import CommonAttrs
 from nexus_constructor.model.node import Node
-from nexus_constructor.model.value_type import ValueType
+from nexus_constructor.model.value_type import ValueType, VALUE_TYPE
+
+
+def _get_human_readable_type(new_value: Any):
```
rerpha, line 10: Use composition here and contain attributes within `Node` and all of its helpers rather than using inheritance. Then we can bin this code off and require the dtype to be completely required rather than guessing it from the value.

PR #817, opened 2020-07-28
Extract strings into constants wherever possible, use ValueTypes class

### LINKED ISSUE #816: use VALUE_TYPE whenever specifying dtype, and make values match filewriter strings
Currently we are specifying dtype as "String", "Double" etc in places, this is prone to typos and may not be correct in the filewriter json. We should use the values in `VALUE_TYPE` instead. 

we should also remove the strings in as_dict methods like `"name"` and put them in a common place to avoid typos as well.

rerpha:
### Issue

Closes #816

### Description of work

Moving dtypes and human readable types with a map for the two into value_types.py 
Assigning constant values for commonly used properties such as `"name"` or `"values"` and putting them in `CommonAttrs.py` (`CommonKeys` and `CommonAttrs`) to avoid potential typos
Fixing type outputting and using a consistent type which matches the filewriter specification. 

### Acceptance Criteria 

No functional changes besides the NC should now output the correct `type` for a dataset (I.E `"Long"` rather than `"int64"`)
Code should look a bit more consistent now as there is a standard to use from, therefore minimising the chance of the outputted JSON being incorrect. 

### UI tests

Some modified to suit changes, but mostly unit tests that deal with the JSON output. 


tests/json/shape_json.py
```diff
@@ -163,7 +163,7 @@ def off_shape_json() -> Dict:
           "type":"dataset",
           "name":"faces",
           "dataset":{
-            "type":"int32",
+            "type":"Integer",
```
matthew-d-jones, line 166: These type strings were correct before, see filewriter documentation: https://github.com/ess-dmsc/kafka-to-nexus/blob/master/documentation/commands.md

nexus_constructor/model/value_type.py
```diff
@@ -1,20 +1,47 @@
 from typing import Union
 import numpy as np
 
+
+class ValueTypes:
```
matthew-d-jones, line 5: The filewriter expects "double", "int64", "uint32" and so on: https://github.com/ess-dmsc/kafka-to-nexus/blob/master/documentation/commands.md

tests/ui_tests/test_ui_array_dataset_table_widget.py
```diff
@@ -68,7 +68,7 @@ def test_GIVEN_string_array_WHEN_changing_data_type_to_int_THEN_array_rests(
 ):
     array = np.array(["hello" for _ in range(10)])
     array_dataset_table_widget.model.array = array
-    array_dataset_table_widget.model.update_array_dtype(VALUE_TYPE["Integer"])
+    array_dataset_table_widget.model.update_array_dtype(VALUE_TYPE_TO_NP["Integer"])
```
matthew-d-jones, line 71: and here

PR #834, opened 2021-01-08
Implement saving and loading depends_on chains of transformations

### LINKED ISSUE #645: Exceptions when linked transformation is renamed
We are not currently updating the link that links to a transformation when the transformation is renamed. We should do this at the same time as updating any dependent transformations.

### LINKED ISSUE #799: JSON Loading: Depends on

### LINKED ISSUE #837: Transform links always apply to first transform in component
To reproduce error:
- make a component and add two for more transforms to it
- make a second component and add a transform to it
- add a link from the first component to the second
- save to JSON

The first transform of the first component will have its depends_on set to the transform in the second component, and other transforms of first component are lost.

matthew-d-jones:
### Issue

Closes #799 
Closes #645
Closes #837

### Description of work

This is the last big change to restore functionality that the NeXus Constructor had before changing the model from hdf5 to Python objects. After merging this and doing some final checks and documentation updates, the model_change branch can be merged back to master and it leaves the codebase in a suitable state for someone at ESS to take over development.

These changes make saving to and then loading from JSON preserve transformations and their depends_on chain.

The approach to loading:
An identifier called `TransformId` is created for dependencies of components and transformations and stored in a dictionary during loading. The `depends_on` property of each component and transformation is then set the correct `Transformation` object once all transformations have been loaded from file. This was a more simple approach than trying to guarantee that the dependency object is created before the object that depends on it.

Created ticket #835 for saving/loading transformations which are KafkaStreams, this does not need to be in model change milestone as it was also not implemented before the model change. There is also a known issue when saving if no value has been entered for a created transformation #836.

### Acceptance Criteria 

Preferably do a manual test by creating a couple of components and adding transformations to them. Try to check the case that the depends_on chain links between components. Save to JSON, then load from JSON and check that everything looks the same in the treeview and 3D view.


tests/json/test_load_from_json.py
```diff
@@ -311,48 +408,64 @@ def test_GIVEN_component_with_nx_class_WHEN_loading_from_json_THEN_new_model_con
...
+def test_GIVEN_json_with_component_depending_on_transfrom_WHEN_loaded_THEN_component_in_model_contains_transform(
+    json_dict_with_component_and_transform, json_reader
 ):
-    transformation = component_with_transformation.transforms[0]
-    assert transformation == json_reader._get_transformation_by_name(
-        component_with_transformation, transformation.name, "DependentComponentName"
-    )
+    json_reader._load_from_json_dict(json_dict_with_component_and_transform)
+    for component in json_reader.entry.instrument.component_list:
+        if component.name == "test_component":
```
SkyToGround, line 416: This unit test is too likely to "fail silently". You need to make sure (guarantee) that the asserts are reached or fail the test if they are not.

tests/json/test_load_from_json.py
```diff
@@ -311,48 +408,64 @@ def test_GIVEN_component_with_nx_class_WHEN_loading_from_json_THEN_new_model_con
...
+            assert len(component.transforms) == 1
+            assert component.transforms[0].name == "location"
 
 
-def test_GIVEN_no_transformation_with_matching_name_WHEN_finding_transformation_by_name_THEN_warning_message_is_created(
-    json_reader, component_with_transformation
+def test_GIVEN_json_with_component_depending_on_non_existent_transfrom_WHEN_loaded_THEN_warning_is_added(
```
SkyToGround, line 421: Btw, these names are getting out of hand in my opinion. There is such a thing as a name that is too long.

tests/json/test_load_from_json.py
```diff
@@ -311,48 +408,64 @@ def test_GIVEN_component_with_nx_class_WHEN_loading_from_json_THEN_new_model_con
...
 
-    transformation_name = component_with_transformation.transforms[0].name
-    dependent_component_name = "DependentComponentName"
+    # Add depends_on dataset which points to a transformation which does not exist in the JSON
+    json_dict_with_component["children"][0]["children"][0]["children"][0][
+        "children"
+    ].append(depends_on_dataset)
+    json_reader._load_from_json_dict(json_dict_with_component)
+
+    assert len(json_reader.warnings) > 0
```
SkyToGround, line 444: Same thing here, this test will too easily fail silently (to test for what it is supposed to). They way I see it there are two options:
1. Modify the test to check that you get the the warning you expect.
2. Make another test that has the minimum difference required not to trigger the warning and that thus will not produce any warning messages.

I prefer #1 but #2 is acceptable.

tests/json/test_load_from_json.py
```diff
@@ -311,48 +408,64 @@ def test_GIVEN_component_with_nx_class_WHEN_loading_from_json_THEN_new_model_con
...
-        [
-            expected_text in json_reader.warnings[-1]
-            for expected_text in [
-                component_with_transformation.name,
-                transformation_name,
-                dependent_component_name,
-            ]
-        ]
-    )
+        len(json_reader.warnings) > 0
```
SkyToGround, line 462: Ditto about warnings.

tests/json/test_transformation_reader.py
```diff
@@ -68,21 +71,34 @@ def attributes_list(transformation_json):
...
     entry = [transformation_json]
-    return TransformationReader(parent_component, entry)
+    transforms_with_dependencies: Dict[
+        TransformId, Tuple[Transformation, TransformId]
+    ] = {}
+    return TransformationReader(parent_component, entry, transforms_with_dependencies)
 
 
-def test_GIVEN_transformation_in_attributes_WHEN_checking_for_transformation_THEN_contains_transformations_returns_true(
+def test_GIVEN_nxtransformation_in_attributes_WHEN_checking_for_transformations_THEN_contains_transformations_returns_true(
```
SkyToGround, line 88: This is ridiculous given that the test is two lines.

mattclarke (approved):
A few minor things, but other than that I think it is good to go.

SkyToGround (changes requested):
There is a bug in this implementation wherein if you try to delete a transformation that has been restored/loaded from file, the application crashes. This needs to be fixed. I can send you an example file if need be.

Otherwise, I see nothing weird in the application as such. Feel free to ignore my comments about the ridiculous unit test names (I know you will).
However, I really think you should fix the unit tests I have commented on as I think they are not really testing what they are supposed to be testing.

tests/json/test_load_from_json.py
```diff
@@ -311,48 +408,64 @@ def test_GIVEN_component_with_nx_class_WHEN_loading_from_json_THEN_new_model_con
     assert json_reader.entry.instrument.component_list[1].nx_class == component_class
 
 
-def test_GIVEN_transformation_with_matching_name_WHEN_finding_transformation_by_name_THEN_transformation_is_returned(
-    json_reader, component_with_transformation
+def test_GIVEN_json_with_component_depending_on_transfrom_WHEN_loaded_THEN_component_in_model_contains_transform(
```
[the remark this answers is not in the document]
matthew-d-jones: Corrected

nexus_constructor/json/transformation_reader.py
```diff
@@ -48,17 +50,38 @@ def _create_transformation_dataset(
     return Dataset(name, size=[1], type=dtype, values=angle_or_magnitude,)
 
 
+def get_component_and_transform_name(depends_on_string: str):
+    # The following extraction of the component name and transformation name makes the assumption
+    # that the transformation lives in a component and nowhere else in the file, this is safe assuming
+    # the JSON was created by the NeXus Constructor
+    depends_on_path = depends_on_string.split("/")
+    dependency_component_name = depends_on_path[-3]
+    # [-2] is the NXtransformations group
```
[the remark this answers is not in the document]
matthew-d-jones: Clarified

nexus_constructor/json/transformation_reader.py
```diff
@@ -217,10 +238,14 @@ def _create_transformations(self, json_transformations: list):
...
-            depends_on = self._find_attribute_in_list(
-                CommonAttrs.DEPENDS_ON, name, attributes
+            # This attribute is allowed to be missing, missing is equivalent to the value "." which means
+            # depends on origin (end of dependency chain)
+            depends_on = _find_attribute_from_list_or_dict(
+                CommonAttrs.DEPENDS_ON, attributes
             )
 
+            # TODO handle cases that "values" is an NXlog, KafkaStream etc
+
```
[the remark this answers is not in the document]
matthew-d-jones: Added

nexus_constructor/component_tree_model.py
```diff
@@ -243,6 +241,14 @@ def _get_target_position(
         target_pos,
         transformation_list,
     ):
+        """
+        @param parent_component: component to add transformation to
```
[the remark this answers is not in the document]
matthew-d-jones: Fixed

tests/json/test_load_from_json.py
matthew-d-jones: Fixed, thanks for pointing out

tests/json/test_load_from_json.py
matthew-d-jones: I've tried to address this by swapping warning strings out for warning objects which can be type checked in the tests. Hopefully this is an acceptable compromise between the test being too lax and too brittle.

Issue #839, opened 2021-01-12
Error dialog when loading JSON


mattclarke:
When loading a file with a few cylindrical detectors I get a pop-up which says:
![image](https://user-images.githubusercontent.com/3032495/104312320-59b59280-54ce-11eb-8dc3-7cc50a84b134.png)

Can be reproduced by just creating a single cylindrical detector of height 1 and radius 1, then saving and reloading.

Note: despite the error, everything seems okay.

kmurica:
Fixed and merged to main.

PR #842, opened 2021-02-24
changing datatype check to check for integers and not floats

kmurica:
### Issue

https://github.com/ess-dmsc/nexus-constructor/issues/839

### Description of work

Erroneous warning message pops up due to datatype check done for float type rather than integer for vertices.

### Acceptance Criteria 

Test if the erroneous warning message still persists.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 19.10b0
     hooks:
     - id: black
       language_version: python3
--- a/nexus_constructor/json/shape_reader.py
+++ b/nexus_constructor/json/shape_reader.py
@@ -220,7 +220,7 @@ class ShapeReader:
             return
 
         vertices_dtype = self._find_and_validate_data_type(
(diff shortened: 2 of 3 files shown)


kmurica:
@matthew-d-jones I'm having some issue with my pre-commit here... Not sure why.
It seems to be due to reformatting of _convert_vertices_to_qvector3d

matthew-d-jones:
@kmurica Probably due to a different version of black in the hook to the requirements file. Try pinning the hook to the same version as in the requirements file, like this: https://github.com/ess-dmsc/forwarder/blob/main/.pre-commit-config.yaml#L3

matthew-d-jones (changes requested):
I think the correct fix here is to allow integers and floats.
Users are allowed to enter a non-integer position in the interface. I think the problem occurs when a user chooses a value that happens to be an integer so when the float is saved to JSON and then loaded back, python converts the string to an integer.
`float -> str -> int` rather than `float -> str -> float`

The list is cast to a qvector3d after this point in the code, I assume that works fine from integer or float. If not then you would need to explicitly cast to float first.

matthew-d-jones (approved):
Thanks. Manually tested and works as expected.

PR #846, opened 2021-03-02
Update Getting started README

ebadkamil:
Update the getting started readme since Nexus File Layout tab does not exist in latest version.


mattclarke:
What are these images? Are they supposed to be empty files?

ebadkamil:
I updated the screenshots too.

ebadkamil:
When we set translation to `+2` in z direction `(0, 0, 1)` why the sample is deliberately moved to `2m` in `-ve z` direction?
https://github.com/ess-dmsc/nexus-constructor/blob/b5246e005947dcd6de5ac32fa3e6d8b6badd4605/nexus_constructor/model/transformation.py#L101

Does it have something to do with the direction of beam?
nexus

kmurica:
> When we set translation to `+2` in z direction `(0, 0, 1)` why the sample is deliberately moved to `2m` in `-ve z` direction?
> https://github.com/ess-dmsc/nexus-constructor/blob/b5246e005947dcd6de5ac32fa3e6d8b6badd4605/nexus_constructor/model/transformation.py#L101
> 
> Does it have something to do with the direction of beam? @kmurica
> 
> nexus



> When we set translation to `+2` in z direction `(0, 0, 1)` why the sample is deliberately moved to `2m` in `-ve z` direction?
> https://github.com/ess-dmsc/nexus-constructor/blob/b5246e005947dcd6de5ac32fa3e6d8b6badd4605/nexus_constructor/model/transformation.py#L101
> 
> Does it have something to do with the direction of beam? @kmurica
> 
> nexus

Just to document it here: the passive transformation is actually moving the coordinate system and not the object. In that case, moving the origo 2*(0, 0, 1) direction, would redefine all object coordinates by -2.

getting_started.md
```diff
@@ -22,10 +22,6 @@ The sample component is added by default, and is indicated by the red cube. You
 
 ![](resources/images/NeXusConstructor_001.png)
 
-If you are familiar with the [NeXus format](resources/introduction_to_nexus.md) it may be useful to view the layout of the resulting NeXus file as you modify the template by adding components and other information. You can do this by clicking the "Nexus File Layout" tab in the left pane of the main window.
-
-![](resources/images/Selection_002.png)
```
kmurica, line 27: reference to resources/images/Selection_002.png is deleted in getting_started.md in this line,
does this mean we can delete it from the repository or is it referenced elewhere? I did a quick search and couldn't find any references to it.

ebadkamil: Right. Thanks. I just removed it from the resources,

kmurica (approved):
LGTM.

Issue #722, opened 2020-04-22
Add a "show button labels" option in a "view" menu


rerpha:
We should add a view menu next to "file" containing a "show button labels" tickbox that shows the labels showing what the buttons are for above the component view.

kmurica:
![image](https://user-images.githubusercontent.com/76992076/108848892-d49bbe00-75e1-11eb-91ea-8dc233c65b8d.png)

matthew-d-jones:
Correct.
Tickets #719 and #720 are closely related. I think this ticket is really a suggestion for part of a solution for #720.
It would be worth looking at how other software deals with this situation (toolbar where it is difficult to convey the meaning of buttons with icons alone). Perhaps the ribbon in MS Office software provides some inspiration for how text and buttons can be used together to help?

ebadkamil:
Thanks for the explanation Matt and Kenan. I am assigning this to myself to get started with `nexus_constructor`

PR #847, opened 2021-03-02
722 Show Button labels in ToolBar

ebadkamil:
Try to address the issue https://github.com/ess-dmsc/nexus-constructor/issues/722 .

### DIFF
--- a/nexus_constructor/main_window.py
+++ b/nexus_constructor/main_window.py
@@ -52,13 +52,6 @@ class MainWindow(Ui_MainWindow, QMainWindow):
         # Clear the 3d view when closed
         QApplication.instance().aboutToQuit.connect(self.sceneWidget.delete)
 
-        # self.treemodel = self.widget.findHdf5TreeModel()
-        # self.treemodel.setDatasetDragEnabled(True)
-        # self.treemodel.setFileDropEnabled(True)
-        # self.treemodel.setFileMoveEnabled(True)
-        # self.treemodel.insertH5pyObject(self.model.signals.nexus_file)
-        self.model.signals.file_changed.connect(self.update_nexus_file_structure_view)
-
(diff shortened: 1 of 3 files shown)


ebadkamil:
Most likely we will need better way to put labels and actions together.

mattclarke:
I'm probably not suitable to be the sole reviewer on NC tickets.
Most of my experience with the NC is watching you guys play with it in the interviews :D

If it is aesthetic/minor changes like this PR include @kmurica as well.
If it is digging into the bowels then include both @kmurica and @matthew-d-jones too.

That way we can all learn more and, hopefully, over the next few months we will build up enough NC knowledge that we won't need @matthew-d-jones T.T

kmurica:
Looks nice! One general thing though.
The without text button layout looks like:
image
The with text buttons is:
image
As I have not used the tool much yet, would it be worthwhile trying to get all the buttons in the button panel visible without the dropdown list? That would probably require to enlarge the button panel to fit the labels . That might end up looking a bit ugly...

ebadkamil:
users can expand the left panel using the splitter and then they all fit in.

matthew-d-jones:
I'm not sure if the "zoom component" button does anything currently, if not you could remove it completely for now. That would help a little with the issue Kenan pointed out.

ebadkamil:
if you select a component on the left panel and click on "zoom component" then it actually zooms in to the component. Although one can just use the scroll button for that.

matthew-d-jones:
Oh great, for some reason I thought that wasn't working.

ebadkamil:
I have renamed the actions rather than stretching the panels on triggering the show button labels. This way they all fit in easily,

kmurica (approved):
LGTM.

PR #850, opened 2021-03-10
843 fix json warning bug

### LINKED ISSUE #843: JSONReader.warnings sometimes returns string type
In **nexus_constructor/main_window.py** we have following code:

`            if reader.warnings:
                show_warning_dialog(
                    "\n".join(
                        (json_warning.message for json_warning in reader.warnings)
                    ),
                    "Warnings encountered loading JSON",
                    parent=self,
                )
`

However, the warnings attribute in reader can sometimes contain a pure string element. Accessing the message attribute from a `str` type is of course not valid and an `AttributeError` is thrown. This will lead to JSON files not being loaded correctly.

Fix this issue by making sure that JSONReader always contains appropriate object types in `reader.warnings`, containing a message attribute. Appropriate class can be found in **nexus_constructor/json/json_warnings.py**.

kmurica:
Closes #843

### DIFF
--- a/nexus_constructor/json/json_warnings.py
+++ b/nexus_constructor/json/json_warnings.py
@@ -63,6 +63,10 @@ class JsonWarningsContainer(List[JsonWarning]):
         self.__add_constructor_param_items(kwargs)
 
     def __add_constructor_param_items(self, other_items):
+        """
+        Internal method to add items into the container if they are provided in
+        the constructor call.
+        """
         if other_items:
             if not any(isinstance(item, self._CHECK_TYPES) for item in other_items):
                 raise TypeError(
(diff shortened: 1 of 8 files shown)


kmurica:
@ebadkamil I think this is ready for review.

nexus_constructor/json/json_warnings.py
```diff
@@ -30,5 +40,88 @@ class NXClassAttributeMissing:
...
+    InvalidTransformation,
+    TransformDependencyMissing,
+    NameFieldMissing,
+    NXClassAttributeMissing,
 ]
+
+_T = TypeVar("_T", JsonWarning, "JsonWarningsContainer")
+
+
+class JsonWarningsContainer(List[JsonWarning]):
```
ebadkamil, line 54: Just for curiosity, Why `List[JsonWarning] `and not just `List`?

kmurica: As we discussed in chat, in principle it is OK to use List but it is mostly to be super clear that it is a JsonWarning container. :)

ebadkamil (commented):
LGTM. In case you need some more opinion, feel free to add others as well.

ebadkamil (approved):

Issue #719, opened 2020-04-22
Unclear which icons are greyed out


rerpha:
They are all grey besides the delete button, so we should either add some colour or raise the buttons that are currently selectable.

rerpha:
related to #720 as that also revolves around the icons being changed.

kmurica:
If colors and icons are kept as is in https://github.com/ess-dmsc/nexus-constructor/issues/720 (i.e. grey), I believe simply raising the buttons should be a sufficient approach for the user to distinguish active and inactive buttons.

ebadkamil:
Raising the button issue was already implemented  in #851 . However adding colors to buttons should remain as an open issue  as suggested by @matthew-d-jones . So not closing this issue as of now.

PR #851, opened 2021-03-10
719 Raise buttons which are currently selectable

ebadkamil:
Closes issue #719

### DIFF
--- a/nexus_constructor/treeview_utils.py
+++ b/nexus_constructor/treeview_utils.py
@@ -137,6 +137,15 @@ def set_button_states(
 
 
 def set_enabled_and_raise(action: QAction, value: bool):
+    """Disable or enable the action and autoRaise the associated QToolButton.
+
+    Parameters
+    ----------
+        action: QAction
+            The action to enable or disable
+        value: bool
+            True to enable action and raise associated QToolButton.
+    """
     action.setEnabled(value)
     for widget in action.associatedWidgets():
         if isinstance(widget, QToolButton):
@@ -159,12 +168,8 @@ def handle_number_of_items_selected_is_not_one(
     :param create_link_action: The action for creating a link.
     :param zoom_action: The action for zooming on a component.
     """
-    set_enabled_and_raise(delete_action, False)
-    set_enabled_and_raise(new_rotation_action, False)
-    set_enabled_and_raise(new_translation_action, False)
-    set_enabled_and_raise(create_link_action, False)
-    set_enabled_and_raise(zoom_action, False)
-    set_enabled_and_raise(edit_component_action, False)
+    for action in vars().values():
+        set_enabled_and_raise(action, False)
 
 
 def expand_transformation_list(
--- a/nexus_constructor/treeview_utils.py
+++ b/nexus_constructor/treeview_utils.py
@@ -3,7 +3,15 @@ import sys
 
 from PySide2.QtCore import QModelIndex, Qt
 from PySide2.QtGui import QColor, QIcon
-from PySide2.QtWidgets import QAction, QFrame, QLabel, QToolBar, QTreeView, QWidget
+from PySide2.QtWidgets import (
+    QAction,
+    QFrame,
+    QLabel,
+    QToolBar,
+    QToolButton,
+    QTreeView,
+    QWidget,
+)
 
 from nexus_constructor.common_attrs import TransformationType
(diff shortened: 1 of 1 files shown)


nexus_constructor/treeview_utils.py
```diff
@@ -82,41 +90,57 @@ def set_button_states(
...
         else:
-            create_link_action.setEnabled(False)
+            set_enabled_and_raise(create_link_action, False)
+
+
+def set_enabled_and_raise(action: QAction, value: bool):
+    action.setEnabled(value)
+    for widget in action.associatedWidgets():
+        if isinstance(widget, QToolButton):
+            widget.setAutoRaise(not value)
```
ebadkamil, line 143: Mac issue observed on @kmurica laptop. Issue described here

https://doc.qt.io/qt-5/qtoolbutton.html#autoRaise-prop

kmurica: I will have a look if there is a relatively easy fix to this issue for mac. Otherwise we might need to create a follow up ticket (?).

kmurica: After some investigation, the problem is not easily fixable:
https://stackoverflow.com/questions/11803260/qtoolbutton-autoraise-not-working-for-mac-osx
@matthew-d-jones The fix requires stylesheets... I am not sure if we should do the fix in this issue, create a new one or just accept it does not work for mac and live with it? Any advice?

umithardal: Can we fix if-else craziness here? May be a dictionary approach could save one-two logic calls. Also, `set_enabled_and_raise` called a lot and in some cases it just differs by one positional argument. A for loop through a list may be better.

nexus_constructor/treeview_utils.py
```diff
@@ -134,11 +159,12 @@ def handle_number_of_items_selected_is_not_one(
...
-    new_rotation_action.setEnabled(False)
-    new_translation_action.setEnabled(False)
-    create_link_action.setEnabled(False)
-    zoom_action.setEnabled(False)
+    set_enabled_and_raise(delete_action, False)
+    set_enabled_and_raise(new_rotation_action, False)
+    set_enabled_and_raise(new_translation_action, False)
+    set_enabled_and_raise(create_link_action, False)
+    set_enabled_and_raise(zoom_action, False)
+    set_enabled_and_raise(edit_component_action, False)
```
umithardal, line 167: Again, 6 lines of same function call. And all of the cases second argument is the same.

nexus_constructor/treeview_utils.py
```diff
@@ -82,41 +90,57 @@ def set_button_states(
...
+            set_enabled_and_raise(
+                create_link_action,
+                not selected_object.parent_component.transforms.has_link,
             )
         else:
-            create_link_action.setEnabled(False)
+            set_enabled_and_raise(create_link_action, False)
+
+
+def set_enabled_and_raise(action: QAction, value: bool):
```
umithardal, line 139: Could you please add docstrings to explain what this function does? Also some comments throughout the code would be helpful to maintain it.

nexus_constructor/treeview_utils.py
ebadkamil: good point. This is fixed now,

nexus_constructor/treeview_utils.py
ebadkamil: if-elif jugglery is too convoluted here. Will have to think about it and manually test each scenario from the gui,  Let's see.

matthew-d-jones (approved):
Code looks fine to me. I think this is an improvement, but adding some colour to the buttons should remain an open ticket: that would hopefully solve the matter on Mac. Style sheets are probably not worthwhile as they can lead to all sorts of other problems with different resolution displays or fighting native OS styling.

What happened to the button labels? I thought that PR was merged but I don't see the labels.

nexus_constructor/treeview_utils.py
umithardal: Sure.

ebadkamil:
@matthew-d-jones you have to click on view in the MenuBar and check show button labels.

matthew-d-jones:
> @matthew-d-jones you have to click on view in the MenuBar and check show button labels.

oh yes, thanks. We should use QSettings so that this persists between application restarts.

PR #852, opened 2021-03-16
ECDC-2404: Save/Load Kafka Stream transformation to FW JSON

### LINKED ISSUE #835: Implement saving/loading JSON for transformations which are a KafkaStream
Transformations can be an f142 KafkaStream which the filewriter creates an NXlog from. This is required when the transformation describes a motion axis which moves during an experiment for example. Currently this is not saved to or loaded from JSON by the NeXus Constructor.

ebadkamil:
Ongoing pair programming with @kmurica. Closes #835

### DIFF
--- a/nexus_constructor/model/transformation.py
+++ b/nexus_constructor/model/transformation.py
@@ -12,6 +12,7 @@ from nexus_constructor.common_attrs import (
     TransformationType,
 )
 from nexus_constructor.model.dataset import Dataset
+from nexus_constructor.model.stream import StreamGroup
 from nexus_constructor.model.value_type import ValueTypes
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


nexus_constructor/model/transformation.py
```diff
@@ -165,22 +166,23 @@ def as_dict(self) -> Dict[str, Any]:
...
+            return_dict = {
+                CommonKeys.NAME: self.name,
+                CommonKeys.TYPE: NodeType.DATASET,
+                CommonKeys.VALUES: value if value is not None else [],
+                CommonKeys.DATASET: {
+                    CommonKeys.TYPE: self.type,
+                    CommonKeys.SIZE: self.size,
+                },
+            }
+        elif isinstance(self.values, StreamGroup):
```
ebadkamil, line 178: Line 164: `as_dict()` method is called twice when we add two translations without kafkaStream. However if we add two translations both with kafkaStream, this method get called only once. Something fishy?

PR #853, opened 2021-03-17
ECDC-2397: Change toolbar button background color if enabled

### LINKED ISSUE #719: Unclear which icons are greyed out

ebadkamil:
With #851 and current PR, we can close #719


kmurica:
Looks good to me. On Mac it looks like follows:
image
Maybe not as beautiful as on Linux, but it works. :)

kmurica (approved):
LGTM

ebadkamil:
Thank you

PR #857, opened 2021-05-17
Fix minor UI issues

SkyToGround:
### Description of work

* Default to use the `double` type instead of `byte`.
* Avoid hiding windows by mistake.
* Add shortcuts for opening and closing.
* Move the "Delete" button to one side in the toolbar.

I wanted to fix some float/int parsing inconsistencies as well but determined after a few hours of working on it that this would take too much time to do right now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,6 @@ numpy
 pytest-qt
 confluent-kafka
 ess-streaming-data-types >= 0.9.5
-babel
 
 
 # PySide2 has had some issues with new versions and packaging so we will pin it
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -10,6 +10,7 @@ numpy
 pytest-qt
 confluent-kafka
 ess-streaming-data-types >= 0.9.5
+babel
```
kmurica, line 13: I don't see that we are using this anywhere here or am I just blind? :)

SkyToGround: Ah, that's true. Its from my attempt at fixing/improving parsing of text field values. I will remove it.
```suggestion
```

kmurica (approved):
Looks good to me. I like the shortcuts. ;)

PR #858, opened 2021-06-01
Various minor bug and ui fixes.

SkyToGround:
### Description of work

* Make sure that "*.json" is last in the file name.
* If we fail to create the JSON-string, do not delete an existing file.
* Do not crash on typing (e.g. "m/s") in a unit field.
* Make sure that we set a locale.
* Fix usage of "tdct" module.

### DIFF
--- a/nexus-constructor.py
+++ b/nexus-constructor.py
@@ -6,6 +6,7 @@ import argparse
 import logging
 import os
 import sys
+import locale
 
 from PySide2 import QtCore
 from PySide2.QtGui import QIcon
@@ -22,6 +23,10 @@ else:
     root_dir = os.path.dirname(os.path.realpath(__file__))
 
 if __name__ == "__main__":
+    if locale.getlocale()[0] is None:
+        used_locale = "en_GB.UTF-8"
+        locale.setlocale(locale.LC_ALL, used_locale)
+        print(f"Unable to determine the system locale, using the default ({used_locale}).")
     parser = argparse.ArgumentParser(description="Nexus Constructor")
     if "help" in parser.parse_args():
         exit(0)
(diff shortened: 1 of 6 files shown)


nexus_constructor/main_window.py
```diff
@@ -156,8 +156,11 @@ def save_to_filewriter_json(self):
         filename = file_dialog(True, "Save Filewriter JSON File", JSON_FILE_TYPES)
 
         if filename:
+            if not (len(filename) > 5 and filename[-5:].lower() == ".json"):
```
ebadkamil, line 159:
```suggestion
            if not filename.endswith(".json"):
```

nexus-constructor.py
```diff
@@ -22,6 +23,12 @@
     root_dir = os.path.dirname(os.path.realpath(__file__))
 
 if __name__ == "__main__":
+    if locale.getlocale()[0] is None:
+        used_locale = "en_GB.UTF-8"
```
ebadkamil, line 27: setting the locale here is a good idea. Just for curiosity, did you face an issue with opening file or something?

ebadkamil:
LGTM.

ebadkamil (approved):

nexus_constructor/main_window.py
kmurica:
```suggestion
            if not filename.lower().endswith(".json"):
```

kmurica: To be on the safe side. The string is immutable anyway so filename should not be impacted.

kmurica (approved):
Fix the file format check. Otherwise LGTM and you can merge this once it has been fixed and tests pass.

nexus-constructor.py
SkyToGround: This actually came from another fix that did not make it into this PR but I thought that I might as well put the set-locale-code here anyway.

PR #863, opened 2021-11-10
ECDC-2435: Enable live update of transformations

umithardal:
### Issue

ECDC-2435

### Description of work
This PR enables live update of value changes in transformations view. In addition some code clear up and enhancement has been done.

Easiest way to see if changes work:

- Change the `value` field next to `Scalar dataset` (under `Distance` if it is a `Translation`, or under `Angle` if it is a `Rotation`. `Vale to use in 3D view` should change simultaneously. Before the changes in this PR, you have to a an `Enter` to update `3D view` value.

### DIFF
--- a/nexus_constructor/component_tree_view.py
+++ b/nexus_constructor/component_tree_view.py
@@ -93,7 +93,7 @@ class ComponentEditorDelegate(QStyledItemDelegate):
     def setModelData(
         self, editorWidget: QWidget, model: QAbstractItemModel, index: QModelIndex
     ):
-        editorWidget.transformation_frame.saveChanges()
+        editorWidget.transformation_frame.save_all_changes()
 
     def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
         model = index.model()
--- a/nexus_constructor/transformation_view.py
+++ b/nexus_constructor/transformation_view.py
@@ -11,7 +11,7 @@ from nexus_constructor.model.model import Model
 from nexus_constructor.model.transformation import Transformation
 from nexus_constructor.unit_utils import METRES, RADIANS
 from ui.link import Ui_Link
-from ui.transformation import Ui_Transformation
+from ui.transformation import UiTransformation
 
 if TYPE_CHECKING:
     from nexus_constructor.model.value_type import ValueType  # noqa: F401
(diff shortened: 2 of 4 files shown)


zjttoefs:
I can see the "Value to use in 3D" change in sync. But the 3D view does not update anymore.

zjttoefs (changes requested):
The 3D view needs to update, obviously. Maybe some of the deleted emits were needed.

There are some previously static methods appear no longer to be static and some comments on those changes would be good to see. In general I would prefer the separate the renaming and reordering from the functional changes into different PRs.

umithardal:
> The 3D view needs to update, obviously. Maybe some of the deleted emits were needed.
> 
> There are some previously static methods appear no longer to be static and some comments on those changes would be good to see. In general I would prefer the separate the renaming and reordering from the functional changes into different PRs.

Now, the description of the problem unfortunately misses the detail. It states that once user changes the focus, the new values takes place. However, until you press `Enter` nothing changes in the 3D view.

With the latest commit I re-added the `emit` signals. You can now see the 3D view updates simultaneously with changes without the need of changing focus or pressing any key. The trade off is we have to emit each time there is change in one of the relevant fields.  I miss calculated the signal-slot behaviour of PYQT, and thus wrongfully removed emits. Thanks for the review and catch.

On the other side of the coin; I had a really hard time understanding the code, and could not solve the problem. So, what I did is to implement a proper `__init__` constructor so that attribute definitions are in proper place. I then refactored to make things readable. With these changes, we only need to create an instance of the object, rather than creating it and calling a method of it.

Only thing that I changed which can be considered as rather a fix is the removal of the inheritance of builtin Python `Object` from the definition of the class. Which was a requirement in Python 2 but as of Python 3 it is not.

Other changes in names basically follows PEP.

mattclarke:
To reinforce what Tobias said.
As a group, we are all guilty of combining refactoring/reformatting with behaviour changes - this needs to stop.
It can make it hard for the reviewer to work out what has changed.
If we do the refactoring/reformatting in a separate PR then it can be reviewed in 30 seconds.
I'll bring this up in the meeting today.

mattclarke:
> Now, the description of the problem unfortunately misses the detail. It states that once user changes the focus, the new values takes place. However, until you press Enter nothing changes in the 3D view.

The description is a starting point. Unfortunately it is unlikely to contain all the information you need to solve the problem especially when it was written by someone intimate with the code base.
This is your cue to explore the details and make sure you understand the problem before getting started.

mattclarke:
> I can see the "Value to use in 3D" change in sync. But the 3D view does not update anymore.

These things happen to all of us from time to time, so don't beat yourself up about it, but take it as a reminder to make sure you test things sufficiently before it gets to review.

mattclarke:
I propose we close this PR and resubmit it as two PRs, one reformatting/refactoring and then one with the functional change.

PR #864, opened 2021-11-10
ECDC-2404: Save/Load kafka Stream to/from JSON

ebadkamil:
* Extends the save/load transformations to json when the values are Streams. Until now only static DataSet could be saved/loaded.
* Only show kafka stream of type f142 in Transformation panel.


kmurica (approved):
Works. LGTM

kmurica:
Just resolve conflicts and it should be fine.

kmurica (approved):
Still works fine. Tests pass. Code is OK. Merge time.

PR #866, opened 2021-11-11
ECDC - 2435: Enable live update

umithardal:
### Issue

ECDC-2435

### Description of work

Enables live update


mattclarke:
This is looking good, but one thing: if I change the type from `Scalar dataset` to `Kafka stream` which enables the `Value to use in 3D view` spinbox then changing that value doesn't update until I click away.

Screenshot 2021-11-12 at 08 29 35

I going to merge this PR as is, so @umithardal could you make a new ticket for the follow-up work for `Value to use in 3D view` please

mattclarke (approved):

PR #867, opened 2021-11-11
ECDC-2435: Tidy up transformation view

umithardal:
### Issue

ECDC-2435 continuation for non-functional changes

### Description of work

Refactors, fixes and clears up.


ui/transformation.py
```diff
@@ -18,46 +18,35 @@
...
+    def __init__(self, Transformation: "EditTransformation"):
         Transformation.setObjectName("Transformation")
         Transformation.resize(361, 171)
-        self.frame_layout = QVBoxLayout(Transformation)
-        self.frame_layout.setContentsMargins(4, 4, 4, 4)
+
         self.main_layout = QVBoxLayout()
         self.main_layout.setSpacing(4)
 
+        self.frame_layout = QVBoxLayout(Transformation)
```
mattclarke, line 29: All this gui stuff should be in the setup_ui method.

ui/transformation.py
```diff
@@ -66,47 +55,61 @@ def setupUi(self, Transformation: "EditTransformation"):
             "Value to use in 3D view:", self.value_spinbox
         )
 
-        self.main_layout.addLayout(self.ui_placeholder_layout)
+        self.setup_ui(Transformation)
 
+    def setup_ui(self, transformation):
```
mattclarke, line 60: Though this is the right thing to do, setupUI is used everywhere else in the code.
I suggest we leave it with the old name for now, and then do a PR where we change it everywhere at once.

ui/transformation.py
```diff
@@ -118,10 +121,17 @@ def set_spinbox_ranges(self):
...
-    def retranslateUi(self, Transformation):
-        Transformation.setWindowTitle(
+    @staticmethod
+    def _make_text_bold(label: QLabel):
+        font = label.font()
+        font.setBold(True)
+        label.setFont(font)
+
+    @staticmethod
+    def retranslate_ui(transformation):
```
mattclarke, line 131: Same as above, let's not change the name in this PR, and instead do a "global" replace across the entire code base in one PR.

mattclarke (commented):
It's good, but see comments.

ui/transformation.py
umithardal: In that case class attributes are defined outside of the constructor. I do not like that. Another way of doing this making setupUI private via `_setupUI` but you do not wanna change the name already, so that is not a choice. If you say just move them whatever, I shall do, but these type of details usually bite our ..s afterwards. :)

kmurica:
@umithardal @mattclarke is this in a mergeable state?

umithardal:
> @umithardal @mattclarke is this in a mergeable state?

Nope, I will make a simple commit that will revert the names. Probably after the meeting it will be ready.

umithardal:
@kmurica after CI, it is now mergeable. Could you do that, thanks.

kmurica (approved):
LGTM. Doesn't seem to break anything that I could find. Tests pass. Code looks OK.

PR #868, opened 2021-11-11
Remove size attribute from Dataset

ebadkamil:
### Description of work

* After consulting with @SkyToGround we decided to remove `size` attribute from `Dataset` class, since it is inferred from array by the FileWriter. `type` attribute is made `optional` and default to `double`
* Fixed a bunch of tests.


SkyToGround (approved):
If it passes the tests, merge it.

PR #869, opened 2021-11-11
transformation_type are lower case

zjttoefs:
transformation_types should be lower case. Also field names are lower case by default. This PR addresses both.


mattclarke (approved):
This is the kind of PR I like!

PR #871, opened 2021-11-12
address warnings with python 3.9

zjttoefs:
ECDC-2699

does not address all potential issues, but removes some warnings and should be backwards compatible down to python 3.3


kmurica (approved):
LGTM. Unit tests succeed.

PR #872, opened 2021-11-15
ECDC-2680: proposal id placeholder

mattclarke:
User can define a proposal id or select to use the placeholder.
If no value specified then nothing gets added to the JSON.

When JSON is loaded the GUI reflects the contents regarding the proposal id.

I've added the GUI bits in a place that isn't too offensive, ultimately we may want to move them when we have the tree-view.


kmurica:
Seems to work nicely! Although I will need to do some changes later on in another ticket: https://jira.esss.lu.se/browse/ECDC-2726 
Should not affect the overall function though.

kmurica (approved):
Looks good to me. Works as intended. Unit tests succeed locally.

PR #873, opened 2021-11-15
ECDC-2374: Deal with units in transformation

ebadkamil:
### Description of work

* Use units in translation/rotation fieldwidget for the 3D view scaling.


kmurica (approved):
LGTM. Seems to work. Only annoying thing is the log:
INFO:root:Unit input d has wrong type. Expected something that could be converted to radians.
INFO:root:Unit input de has wrong type. Expected something that could be converted to radians.
But that was affected from an earlier PR.

PR #878, opened 2021-11-17
ECDC-2726_remove_old_dataset_class

kmurica:
### Issue

ECDC-2726

### Description of work

Changing code to use the dataset in stream.py instead.


ebadkamil (approved):
LGTM

PR #881, opened 2021-11-17
Ecdc 2730 disk chopper orientation

zjttoefs:
### Issue

ECDC-2730


kmurica (approved):
Works for me and seems more correct with the nexus definition.

PR #886, opened 2021-11-19
ECDC-2697: Allow users to be defined in the GUI

mattclarke:
Adds a dialog to allow adding users to the JSON.
It saves to the JSON and can be read back.

The part of the ticket for adding the placeholder will come in another PR.

### DIFF
--- a/nexus_constructor/json/load_from_json.py
+++ b/nexus_constructor/json/load_from_json.py
@@ -284,6 +284,8 @@ class JSONReader:
                 and parent_node.nx_class == "NXentry"
             ):
                 self.model.entry[nexus_object.name] = nexus_object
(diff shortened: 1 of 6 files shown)


nexus_constructor/json/load_from_json.py
```diff
@@ -284,6 +284,8 @@ def _read_json_object(self, json_object: Dict, parent_node: Group = None):
                 and parent_node.nx_class == "NXentry"
             ):
                 self.model.entry[nexus_object.name] = nexus_object
+            if isinstance(nexus_object, Group) and nexus_object.nx_class == "NXuser":
```
mattclarke, line 287: Bit of a hack, but as this is the old code I think we can put up with it.

kmurica (approved):
LGTM. Works fine and unit tests pass. Code looks good.

PR #893, opened 2021-11-29
ECDC-2723: Visualise neutron source

ebadkamil:
### Issue

* Closes ECDC-2723.

### Description of work

* Add visualization for neutron source, same as the one we see in Gnomon. Except the neutron animations is removed and neutrons are added to random positions inside a cylinder  mesh.
* Fixes Bug related to sample color which is supposed to red. When one changes the name of sample, the color becomes black.


nexus_constructor/component_type.py
```diff
@@ -101,6 +101,8 @@
 }
 
 CHOPPER_CLASS_NAME = "NXdisk_chopper"
+SOURCE_CLASS_NAME = "NXsource"
+SAMPLE_CLASS_NAME = "NXsample"
```
kmurica, line 105: Could you check for occurrences of "NXsample" in the repo (except in component_type.py) and replace with SAMPLE_CLASS_NAME.

nexus_constructor/instrument_view/instrument_view.py
```diff
@@ -211,24 +215,30 @@ def add_component(
...
+        if nx_class == SOURCE_CLASS_NAME:
+            entity = NeutronSource(self.component_root_entity)
+            self.component_entities[name] = entity
+        else:
+            mesh = OffMesh(geometry.off_geometry, self.component_root_entity, positions)
+            material = create_material(
+                MATERIAL_COLORS.get(nx_class, QColor("black")),
+                QColor("grey"),
+                self.component_root_entity,
+                MATERIAL_ALPHA.get(nx_class, None),
```
kmurica, line 227: MATERIAL_ALPHA.get(nx_class) should be enough.

nexus_constructor/instrument_view/instrument_view.py
```diff
@@ -211,24 +215,30 @@ def add_component(
...
+                [mesh, material], self.component_root_entity
+            )
 
     def get_entity(self, component_name: str) -> Qt3DCore.QEntity:
         """
         Obtain the entity from the InstrumentView based on its name.
         """
         try:
-            return self.component_entities[component_name]
+            component = self.component_entities[component_name]
```
kmurica, line 238: Hmm.. I would prefer if the component_entities actually only contain just that. Do you really need the reference here to be of type NeutronSource? Otherwise these isinstance checks will grow in the repo.

nexus_constructor/main_window.py
```diff
@@ -129,7 +129,9 @@ def _update_views(self):
     def _update_3d_view_with_component_shapes(self):
         for component in self.model.entry.instrument.component_list:
             shape, positions = component.shape
-            self.sceneWidget.add_component(component.name, shape, positions)
+            self.sceneWidget.add_component(
+                component.name, component.nx_class, shape, positions
```
kmurica, line 133: as python, excluding primitive types, sends everything by reference, we could just have add_component(component: Component) and extract everything we need in add_component. In that way we do not need to update everything everytime we need something from the Component class. What do you think?

nexus_constructor/instrument_view/qentity_utils.py
```diff
@@ -49,3 +61,116 @@ def create_qentity(
     for component in components:
         entity.addComponent(component)
     return entity
+
+
+class NeutronSource:
```
kmurica, line 66: maybe having this in the geometry package would be better (see slit and disk_chopper).

nexus_constructor/instrument_view/qentity_utils.py
```diff
@@ -49,3 +61,116 @@ def create_qentity(
...
+        for i in range(self._num_neutrons):
+            mesh = Qt3DExtras.QSphereMesh(self.root_entity)
+            mesh.setRadius(neutron_radius)
+
+            transform = Qt3DCore.QTransform(self.root_entity)
+            transform.setMatrix(
+                self._get_sphere_transformation_matrix(self._neutron_offsets[i])
+            )
+            neutron_material = create_material(
+                QColor("black"), QColor("grey"), self.root_entity
```
kmurica, line 137: some neutrons are blue and some are black when I add the source to the instrument view.

nexus_constructor/instrument_view/qentity_utils.py
```diff
@@ -49,3 +61,116 @@ def create_qentity(
...
+        cone_transform.setMatrix(self._get_cylinder_transformatrion_matrix())
+        material = create_material(
+            QColor("blue"), QColor("lightblue"), self.root_entity, alpha=0.5
+        )
+        self._source = create_qentity(
+            [cylinder_mesh, material, cone_transform], self.root_entity
+        )
+        self._setup_neutrons()
+
+    def setParent(self, value=None):
```
kmurica, line 96: better to not call setParent if value is None.

kmurica (changes requested):
Changes according to comments. Let us discuss on Friday.

nexus_constructor/instrument_view/q_components.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def remove_transformation(self, transformation: Qt3DCore.QComponent):
+        for entity in self.entities:
+            entity.removeComponent(transformation)
+
+    def setParent(self, value=None):
+        for entity in self.entities:
+            entity.setParent(value)
+
+    def get_entity(self):
```
kmurica, line 41: I assume there is a reason why it returns the first element in the list? Not super clear to me. Maybe rename the function or add docstring explaining it.

nexus_constructor/instrument_view/q_components.py
```diff
@@ -0,0 +1,168 @@
...
+        super().__init__(root_entity, nx_class)
+        self._mesh = mesh
+
+    def create_entities(self):
+        self.entities.append(
+            create_qentity([self._mesh, self.default_material], self.root_entity)
+        )
+
+    def add_transformation(self, transformation: Qt3DCore.QComponent):
+        self.entities[0].addComponent(transformation)
```
kmurica, line 64: also here maybe add a short docstring. Does it apply transformation to all entities?

kmurica: Ah, it only has one component in the list? it might still be good to go through the list and add the transformation for all entities in the list even if we are pretty sure it will only contain one for all future.

nexus_constructor/instrument_view/q_components.py
```diff
@@ -0,0 +1,168 @@
...
+                [
+                    r * np.cos(theta),
+                    r * np.sin(theta),
+                    np.random.uniform(-height / 2, height / 2),
+                ]
+            )
+        return np.array(offsets)
+
+    @staticmethod
+    def _get_cylinder_transformatrion_matrix() -> QMatrix4x4:
```
kmurica, line 160: spelling mistake of transformation

nexus_constructor/instrument_view/q_components.py
```diff
@@ -0,0 +1,168 @@
...
+        self._num_neutrons = 8
+        self._neutron_offsets = self._generate_random_points_in_cylinder(
+            self._num_neutrons, self._source_radius, self._source_length
+        )
+
+    def create_entities(self):
+        self._create_source()
+        self._setup_neutrons()
+
+    def add_transformation(self, transformation: Qt3DCore.QComponent):
```
kmurica, line 81: I think this function can be simpler. One suggestion, why not to the entities list append tuples where first element is the entity itself and the second element is either _get_cylinder_transformatrion_matrix() or self._get_sphere_transformation_matrix(offset) and then you do not need to protected methods to set the transformations. Also you can get rid of the "if index" thing etc. Does it make sense?

kmurica (changes requested):

kmurica (changes requested):
looks good. Some minor comments.

nexus_constructor/instrument_view/qentity_utils.py
kmurica: Seems to be a mac issue.

nexus_constructor/main_window.py
kmurica: Fix in later PR.

kmurica (approved):
Looks great now. 🗡️

PR #894, opened 2021-11-30
ECDC-2732 transform and is_component in Group class

kmurica:
### Issue

ECDC-2732

### Description of work

Adding member transformation functions to the Group class needed to generate a proper component.


kmurica:
New PR will be created. Do not delete the branch though.

PR #895, opened 2021-12-06
ECDC-2738 Remove component list

kmurica:
### Issue

ECDC-2738

### Description of work

Remove component list from instrument class. This fixes the issue with copies of components when a json is loaded, edited and saved again.

### DIFF
--- a/nexus_constructor/component_tree_model.py
+++ b/nexus_constructor/component_tree_model.py
@@ -83,6 +83,7 @@ class ComponentTreeModel(QAbstractItemModel):
     def add_component(self, new_component: Component):
         self.beginInsertRows(QModelIndex(), len(self.components), len(self.components))
         self.components.append(new_component)
+        self.model.entry.instrument[new_component.name] = new_component
         self.endInsertRows()
 
     def _remove_link(self, index: QModelIndex):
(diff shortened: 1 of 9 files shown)


nexus_constructor/component_tree_model.py
```diff
@@ -26,7 +26,7 @@ class ComponentTreeModel(QAbstractItemModel):
     def __init__(self, model: Model, parent=None):
         super().__init__(parent)
         self.model = model
-        self.components = self.model.entry.instrument.component_list
+        self.components = self.model.entry.instrument.get_components()
 
     def columnCount(self, parent: QModelIndex) -> int:
```
umithardal, line 31: Can we please remove this function? Why do we need a function to return an integer?

umithardal (approved):
As far as I see the changes satisfies what has been asked in the ticket; basically improve on component list.

Other than my comment on a function that returns an integer which is `1` (this is a trick in python to make sure we have a constant, as Python has no constants), I saw nothing wrong. So I approved.

kmurica:
Added issues with inherited functions to ticket https://jira.esss.lu.se/browse/ECDC-2679

PR #897, opened 2022-01-05
ECDC-2737: respect dataset type when writing to JSON

ebadkamil:
### Issue

Close ECDC-2737

### Description of work
- When writing to json, dataset's type is respected.


kmurica (approved):
Code is good. Works and unit tests pass.
LGTM.

PR #898, opened 2022-01-05
ECDC-2746: Error loading amor json

kmurica:
### Issue

Does not close the ticket.

### Description of work

Adding handling for the case when the group lacks a proper nexus class when loading a json and putting a warning.

### Acceptance Criteria 

load the amor json file and see that the warning is there for those cases.

### UI tests

load the amor json file.

### DIFF
--- a/nexus_constructor/model/attributes.py
+++ b/nexus_constructor/model/attributes.py
@@ -27,11 +27,8 @@ class Attributes(list):
         )
 
     def get_attribute_value(self, attribute_name: str):
-        item = _get_item(self, attribute_name)
-        if contains_attribute(attribute_name):
+        if self.contains_attribute(attribute_name):
             return _get_item(self, attribute_name).values
-        return None       
-            return item.values
         return None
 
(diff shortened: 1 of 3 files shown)


nexus_constructor/model/attributes.py
```diff
@@ -27,7 +27,10 @@ def set_attribute_value(
         )
 
     def get_attribute_value(self, attribute_name: str):
-        return _get_item(self, attribute_name).values
+        item = _get_item(self, attribute_name)
+        if item:
```
ebadkamil, line 31:
```suggestion
        if contains_attribute(attribute_name):
            return _get_item(self, attribute_name).values
        return None       
```

ebadkamil: Just to make use of already existing function below.

ebadkamil (approved):
I was able to load amor json file without error. Too many warnings but that's what is expected from a json file which was probably not produced by NC.

LGTM even if you don't take the minor suggestion.

PR #906, opened 2022-01-31
fix bug as PR says

kmurica:
### Issue

Bug fix

### Description of work

Fixed bug where dataset would always revert to array dataset field type.

### Acceptance Criteria 

test that it does not when you load the json test file and edition for example description dataset value.

### UI tests

test according to acceptance criteria


kmurica:
Bug fix is moved to other PR concerning links. Closing this one.

PR #908, opened 2022-02-01
ECDC-2772 new model implementation

kmurica:
### Issue

Closes ECDC-2772

### Description of work

Getting rid of old model and replacing it with the new one.
Removing environment variable USE_NEW_TREE_STRUCT.
Old way should not be an option anymore.

### Acceptance Criteria 

It works as before but should be more flexible.
Adding and editing components/groups/datasets on any level of the nexus structure.

### UI tests

load, save json files. Play around with the functionalities of nexus constructor.


nexus_constructor/add_component_window.py
```diff
@@ -532,18 +543,31 @@ def edit_existing_component(
         :return: The geometry object.
         """
         # remove the previous object from the qt3d view
-        self.parent().sceneWidget.delete_component(self.component_to_edit.name)
-
+        children_copy = deepcopy(self.component_to_edit.children)
+        for child in children_copy:
+            child.parent_node = None
```
ebadkamil, line 548: Why `parent_node` is set to `None` here?

nexus_constructor/add_component_window.py
```diff
@@ -627,3 +651,27 @@ def update_pixel_input_validity(self):
...
+    Adds fields from a list widget to a component.
+    :param component: Component to add the field to.
+    :param fields_widget: The field list widget to extract field information such the name and value of each field.
+    """
+    for i in range(fields_widget.count()):
+        widget = fields_widget.itemWidget(fields_widget.item(i))
+        try:
+            component[widget.name] = widget.value
+            if component_model:
+                component_model.rowCount(component_model.current_nxs_obj[1])
```
ebadkamil, line 669: Do we need this in every iteration of `for` loop?

nexus_constructor/add_component_window.py
```diff
@@ -532,18 +543,31 @@ def edit_existing_component(
...
-        add_fields_to_component(self.component_to_edit, self.fieldsListWidget)
-        self.generate_geometry_model(self.component_to_edit, pixel_data)
+        if isinstance(self.component_to_edit, Component):
+            self.write_pixel_data_to_component(
+                self.component_to_edit, nx_class, pixel_data
+            )
+            self.generate_geometry_model(self.component_to_edit, pixel_data)
+        for child in children_copy:
+            if isinstance(child, Group):
+                self.component_to_edit[child.name] = child
```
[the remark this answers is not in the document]
ebadkamil: Nevermind, I think I got it.

nexus_constructor/json/load_from_json.py
```diff
@@ -293,24 +295,6 @@ def _validate_nx_class(self, name: str, nx_class: str) -> bool:
             return False
         return True
 
-    def _fit_into_model(self):
```
ebadkamil, line 296: Great!! Finally gone

nexus_constructor/model/attributes.py
```diff
@@ -50,6 +53,7 @@ class FieldAttribute:
 
     name = attr.ib(type=str)
     values = attr.ib(type=ValueType, cmp=False)
+    parent_node = attr.ib(type=Attributes, default=None)
```
ebadkamil, line 56: Should `__eq__` also check if `parent_node` are same?

ebadkamil (commented):
Works as expected.

nexus_constructor/model/attributes.py
kmurica: I don't think it is needed. The attributes are only checked with __eq__ for the same parent_node. For example if dataset dataset_1 has 3 attributes, they will only be compared to each other and not to another set of attributes in say dataset_2. That comparison would not be necessary ever.

nexus_constructor/add_component_window.py
kmurica: You are correct, it is not needed. I think that is some leftover from when I was debugging segmentation faults (tried everything :D ).

nexus_constructor/add_component_window.py
kmurica: Nope, just moved it out of the for loop. Just pushed the changed.

nexus_constructor/model/component.py
```diff
@@ -462,13 +453,14 @@ def _create_transformation_vectors_for_pixel_offsets(
             z_offsets = self.get_field_value(Z_PIXEL_OFFSET)
         except AttributeError:
             z_offsets = np.zeros_like(x_offsets)
+        if not isinstance(x_offsets, list):
```
ebadkamil, line 456: May be a small comment here as to why flatten is used. Not very clear as of now.

kmurica: It is old code really, once it is saved to a json it is in a list. However, I think internally it will be a np.array when you define the detector from the UI.

nexus_constructor/treeview_utils.py
```diff
@@ -280,3 +286,5 @@ def get_module_frame(frame: QFrame, model: Model, value: FileWriterModule):
     module_frame = ModuleView(value, frame, model)
     frame.module_frame = module_frame
     frame.layout().addWidget(frame.module_frame)
+    if isinstance(value, Dataset) and value.name in EXCLUDED_PIXEL_GRID:
```
ebadkamil, line 289: I would suggest to not show at all. Disabling it doesn't help too much as one can not see the content because even Edit button will be disabled and array table won't be visible.

PR #909, opened 2022-02-08
ECDC-2371: Add Status indicator decorator

ebadkamil:


kmurica:
Closing PR in its current state.

PR #910, opened 2022-02-18
ECDC-2798 Do not automatically add stream module in NXlog

kmurica:
### Issue

ECDC-2798, ECDC-2783

### Description of work

1. Stream module should NOT be added in NXlog. NXlog can be added as a group now and then add manually a stream module. The reason being that there are other nexus groups that can have stream modules (for example NXevent_data).

2. Adding a proper ADar schema stream widget configuration with array size list configuration (must be in ADar configuration to work properly!).

### Acceptance Criteria 

1. For NXlog: See description.

2. For ADar schema widget: UI unit test passes. User input works and saves a proper json file.

### UI tests

1. See that behaviour seems reasonable and that an NXlog group can be added together with a stream module.

2. Test to add a ADAr stream module in for example Entry and save a json.


kmurica:
Merged ECDC-2783 into this PR as well as I branched out from this and it is semi related.

ebadkamil (approved):
LGTM

PR #913, opened 2022-02-22
ECDC-2800: Bug fix in deleting users in configure users table

ebadkamil:
### Issue

* Closes ECDC-2800


kmurica (approved):
LGTM. Works as intended compared to main branch.

PR #914, opened 2022-02-24
Live update when spin box changed

umithardal:
This resolves the live update issue of `3D View` when the type is not scalar. Solution works for every type but `Array`which generates a bug concerning array dimension. I tried to resolve it with no success. May be one has to do some edit before changing `3D View` for `Array` types. If that is the case, when `Array` type selected, the `3D View` spin box should be disabled until the correct edits have been done. This is out of scope of my knowledge, if suggested I can open a ticket.


ebadkamil (approved):

ebadkamil:
Works as expected. Thanks,

PR #916, opened 2022-03-06
ECDC-2804 NXdetector bug fix regarding streams

kmurica:
### Issue

Closes ECDC-2804

### Description of work

Fixes bug when editing stream modules added directly to a component.
This is actually a more general bug than just for NXdetector. 

Test when adding component with a dataset that once you edit the component and add a stream module directly (without a group), the stream module is added correctly in the treeview (was previously added as an empty dataset).
Editing the component again should display the stream module correctly in the field list in the widget.


ebadkamil (approved):
LGTM

PR #919, opened 2022-03-16
ECDC-2844 Removing the stream in parent of parent in edit group dialogue

kmurica:
### Issue

Closes ECDC-2844 (https://jira.esss.lu.se/browse/ECDC-2844)

### Description of work

Do not display stream module in parent of parent when editing a group containing streams.

### Acceptance Criteria 

No stream module child of group child when editing group.

### UI tests

Create a Group "A" (arbitrary which nx_class).
Put an NXlog group as child of that group and add a stream module to the group.
When editing Group "A", stream module widget should not be visible in edit group window.


SkyToGround (approved):
Seemed to work properly when doing a quick test of the application.

PR #922, opened 2022-03-23
ECDC-2853: Improving and fixing stream widget editing in tree view

kmurica:
### Issue

Closes https://jira.esss.lu.se/browse/ECDC-2853

### Description of work

Fixing bug that made it impossible to edit Kafka streams in the tree view directly on mac.
Improvement to editing capabilities in tree view overall.

### Acceptance Criteria 

Editing Kafka streams directly in tree view should work better and be more flexible than before, in the tree view mode.

### UI tests

Try out by clicking and editing Kafka streams in the tree view instead of component editor.


SkyToGround (approved):
There are still multiple "quirks" with the UI that should be fixed but I think it is probably easier to do that at a later point.

PR #924, opened 2022-03-29
fixing target_index when transformation group is selected

kmurica:
### Issue

Very minor fix to retrieve the correct target index when transformation group is selected when adding transformation.

### Description of work

minor bug fix.


kmurica:
@SkyToGround I can't merge this so need you to merge it for me :/

PR #927, opened 2022-03-31
Ecdc 2874 performance improvements

SkyToGround:
### Description of work

Turns out that initialising the unit registry is super slow.


kmurica (approved):
Very nice! Quite a performance improvement from what I can see!

PR #928, opened 2022-03-31
ECDC-2875 Constrict nx class suggestion when editing group

kmurica:
### Issue

Closes https://jira.esss.lu.se/browse/ECDC-2875

### Description of work

When an existing component or group is edited in the edit group window, it should only suggest nexus classes in the combo box that maintains the object type (i.e. either object type of class Component or Group) using COMPONENT_TYPES and NX_CLASSES sets.


SkyToGround (approved):
I would personally prefer if you used a non-editable combo box for selecting the class and it having an invalid (empty) selection when opening the "create component" window. This is fine for now though so merge if you want.

PR #931, opened 2022-04-07
General bug fixes

SkyToGround:
### Description of work

* Fixed a crashing bug when saving to JSON
*  Improved the verification of dataset names. Needs more work but its a start at least.


kmurica (approved):
LGTM

PR #933, opened 2022-04-19
Various UI improvements

SkyToGround:
### Description of work

* Fixed index creation bug in the tree view.
* Various minor fixes/improvements to the tree view UI.
* Fixed/improved logic for moving the selection when deleting an item in the tree view.


kmurica (approved):
LGTM

PR #938, opened 2022-04-21
ECDC-2872 Limit writer modules

SkyToGround:
### Description of work

* Various


kmurica (approved):
Looks good to me now. Fixed the bugs I have found. There might be more, but for now I think it is fine. Please, merge if it is OK with you. Would be nice to have this rather big PR merged now. :D

PR #950, opened 2022-05-06
Revert the requirements-dev.txt hacks made for the interviews.

mattclarke:
The original commits can just be reverted but this is here just as a reminder to do so.


kmurica:
I have created a separate requirements-jenkins.txt

PR #977, opened 2022-08-15
improving loading off file

kmurica:
### Issue

Speed up the OFF file loading.

### Description of work

Preallocating lists when loading an OFF file.


kmurica:
No improvement can be seen.

PR #986, opened 2022-08-26
ECDC-2373_setting_fixed_locale

hurvan:
### Issue

Issue 1: Closes [ECDC-2373](https://jira.esss.lu.se/browse/ECDC-2373)

Issue 2: When trying to run the program this error comes:

  File "/home/jonas/code/nexus-constructor/nexus-constructor.py", line 90, in <module>
    window.show()
  File "/home/jonas/code/nexus-constructor/nexus-constructor.py", line 47, in show
    self.resize(
TypeError: 'PySide2.QtWidgets.QWidget.resize' called with wrong argument types:
  PySide2.QtWidgets.QWidget.resize(str, str)
Supported signatures:
  PySide2.QtWidgets.QWidget.resize(PySide2.QtCore.QSize)
  PySide2.QtWidgets.QWidget.resize(int, int)
Failed to make context current: OpenGL resources will not be destroyed
Segmentation fault (core dumped)

### Description of work

Sets the C as the default locale to avoid issues from different computers system locale settings. ( "." is the default delimiter instead of sometimes ",")
Previous settings of locale did not seem to impact PyQt, so QLocale was used instead.

There was also a bug in ubuntu 22.04 where it was not possible to start the program due to a typeError which was fixed by converting the string values to INT's.

### Acceptance Criteria 

A fixed locale has been sent. To test it you can try switching your system region and language and run the program and check if the decimal delimiter is "." or ",". It should be ".".

The resize method that is called upon starting the program will now take INT's just like the move method that follows it. To test it you can run the program, change the window size, close the program, start the program again and verify that the window size is the same.

### UI tests

I have not updated the tests since I'm not sure the tests should change the computers system wide region, language and keyboard settings.


kmurica (approved):
Looks good to me!

PR #999, opened 2022-10-05
ECDC-3039: Show group parent docs at start

kmurica:
Shows parent group documentation when:
1. nx_class has not yet been selected (i.e. None)
2. when the nx_class is NXlog or NXevent_data
Shows group documentation otherwise.

Closes ECDC-3039


hurvan (approved):
Looks good to me. Runs on Ubuntu 22.04

PR #1001, opened 2022-10-07
Render settings

hurvan:
I have added a render settings tab where you can toggle if meshes should render as full meshes or convex hulls in the instrument view. There is also a way of changing material colours.

There was also an update on how materials are managed in general. This is now in a global dictionary instead of being hardcoded and spread out throughout the different files.


kmurica (approved):
Looks nice!

PR #1003, opened 2022-10-10
Ecdc 2979 drag drop groups

hurvan:
Added drag-and-drop functionality for moving groups and components in the treeview. This might break the drag and drop functionality from wherefore, in that case, it is probably in the function "mimeData" or "canDropMimeData" that the issue lies.


kmurica (approved):
Looks good!

PR #1006, opened 2022-10-12
Fixed small bug that caused the program to render removed geometry

hurvan:
Changed "name" to "absolute_path" since the dictionary with component_entities has been changed to used "absolute_path" as keys.

Try adding a geometry, then editing it to ensure that the old geometry is removed and the new one is used.


kmurica (approved):
Works as should.
LGTM

PR #1007, opened 2022-10-17
 ECDC-3038: Adding offset to transformations

kmurica:
### Issue

Closes ECDC-3038

### Description of work

Added possibility to set an offset to transformations.

### Acceptance Criteria 

It should work as intended both in tree view and in instrument view.
The offset is applied first and uses the same vector as the rest of the transformation.

### UI tests

Test by adding transformation to a component, editing the offset (see that both widget and 3D view updates).
Save and load JSON should work as well.


hurvan (approved):
Works on Ubuntu 22.04!

PR #1010, opened 2022-11-08
ECDC-3189: Adding se00 writer module

kmurica:
### Issue

Closes https://jira.esss.lu.se/browse/ECDC-3189

### Description of work

 Adding se00 writer module.


kmurica:
Haven't seen any problems with this in my testing so I will just merge it now.

PR #1013, opened 2022-12-14
ECDC-3262: Enable NeXus skeleton file creation

kmurica:
### Issue

Closes [ECDC-3262](https://jira.esss.lu.se/browse/ECDC-3262)

### Description of work

This PR enables the possibility to save a simple NeXus file for further investigation (i.e. scipp and similar).

### Acceptance Criteria 

It should contain geometry, transformations and overall tree structure.
Should include static datasets and links but not stream modules.
We could create simulated data for the different schemas and save that, but I think that would warrant a separate ticket.


hurvan (approved):
Tested writing a nexus file and open it in python, and it seems to work fine!
So looks good to me.

PR #1014, opened 2022-12-15
ECDC-3265 fixing se00 type

hurvan:
### Issue

SE00 was saved to the json file without any dtype.

### Description of work

Added the option of choosing dtype and made sure that it is being written to the json file

### Acceptance Criteria 

Make sure that the written SE00 has a dtype


kmurica (approved):
LGTM.

PR #1023, opened 2023-02-24
added 3.11

mattclarke:


mattclarke:
Needs to wait for open3d to support 3.11

PR #1024, opened 2023-03-16
ECDC-3354 populate value_units field in f14x filewriter modules

danesss:
### Issue

Nexus files lacked `units` properties for NXlog datasets from Kafka streams.
https://jira.esss.lu.se/browse/ECDC-3354

It was actually possible to setup units in the Nexus Constructor GUI, however it was confusing because the "units" field in the component tree was completely ignored, and the units had to be set by clicking "View" and setting `Value units` in the f144 specific widget.

### Description of work

I opted for removing one of the two GUI component dedicated to setup units.
I chose to keep the one in the component tree. This makes the user experience more consistent, as all dataset types will configure units the same way. The trade-off is that the model (`StreamModule`) has to be updated from a parent widget (`FieldWidget`) instead of the widget that directly wraps the model ( `StreamFieldsWidget`).

### Acceptance Criteria 

* Is the removal of `Value units` in the f144 editing widget is a desirable GUI change? (to favour a consistent definition of units across all dataset types).
* Is the method to update the model from a parent widget robust enough?
* Are the updated UI tests adequate?

### UI tests

* 3 tests were replaced with a single test that replicates the GUI flow used to update units.


hurvan (approved):
LGTM

PR #1026, opened 2023-03-23
ECDC-3354 missing value_units field in f14x filewriter modules

danesss:
### Issue

Nexus files lacked `units` properties for NXlog datasets from Kafka streams.
https://jira.esss.lu.se/browse/ECDC-3354

It was actually possible to setup units in the Nexus Constructor GUI, however it was confusing because the "units" field in the component tree was completely ignored, and the units had to be set by clicking "View" and setting `Value units` in the f144 specific widget.

### Description of work

There were 3 places where units could be configured. This commit makes one of them editable and the other two read-only,

### Acceptance Criteria 

* Is the new user experience adequate?
* Are the updated UI tests adequate?

### UI tests

* A test that replicates the GUI flow used to update units has been added.


hurvan (approved):
Looks good to me! I tested it, and it behaves more logically than before.

PR #1032, opened 2023-06-30
 ECDC-3503-Fix-for-MacOS-dark-mode-breaks-presentation-in-Linux

ggoneiESS:
### Issue

Closes  ECDC-3503-Fix-for-MacOS-dark-mode-breaks-presentation-in-Linux 

Ultimately this will be better fixed with a global stylesheet, but outside of the scope of this ticket

### Description of work

Added If on fusion presentation, which was the GUI change implemented

### Acceptance Criteria 

GUI is usable in Linux

### UI tests

No changes (automated testing did not catch this and passed with validation)

- [danesss] Nominate for code review

### DIFF
--- a/nexus-constructor.py
+++ b/nexus-constructor.py
@@ -75,7 +75,8 @@ if __name__ == "__main__":
     if "help" in parser.parse_args():
         exit(0)
     logging.basicConfig(level=logging.INFO)
-    QApplication.setStyle("Fusion")
+    if sys.platform == "darwin":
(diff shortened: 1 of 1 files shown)


ggoneiESS:
yes you're right, I missed that in the comparison of the last commits.  There are also two tests that needed altering.  I have put these all in conditions.  I can see no other relevant changes that were done by me previously, and it seems 'Darwin' should only catch macOSes.  I've changed that here in an if and suggest deleting the other branch.

Review: Check for correct GUI behaviour when running.

PR #1039, opened 2023-09-21
Change to orbit camera as Constructor does not need to run Doom

ggoneiESS:
### Issue
The viewer for the Constructor uses a first person shooter-style camera.  Since we only need to process JSON that creates instruments etc., rather than play DOOM, this is stupid.  We use Orbit instead.

### Description of work
Changed to QOrbitCameraController class from Qt3DExtras

### Acceptance Criteria 
Easier to use than before

### UI tests
N/A


amues:
The code is fine and I tested it manually, it seems to work fine. As I mentioned, I believe it is a good idea to check with the rest of the group before merging this, as it's a change to how the nexus-constructor works.

amues (approved):

PR #1044, opened 2023-10-30
Ecdc 3633 add component window to widget

ggoneiESS:
### Issue
We want to move the add component window from being a window to being a widget.  The behaviour should be similar, but NOT block the navigation of the tree or other components.  The behaviour should otherwise be the same.

### Description of work
QDialog is a child of QWidget, so that is simple to change.  However, the window object is normally deleted, and that is something we want to avoid, so instead we use the hidden attribute.  The height of the window changes based on what it is populated with.

### Acceptance Criteria 
Files associated with the 'add component' window

### UI tests
All tests still valid

### DIFF
--- a/nexus_constructor/add_component_window.py
+++ b/nexus_constructor/add_component_window.py
@@ -508,9 +508,6 @@ class AddComponentDialog(Ui_AddComponentDialog):
         self.fileLineEdit.setText(filename)
 
     def height_reset(self, height):
-#        minimumHeight = self.sizeHint().height()
-        print(self.shapeOptionsBox.sizeHint().height() + self.fieldsBox.sizeHint().height() + self.unitsbox.sizeHint().height() + height)
-        print(self.shapeOptionsBox.sizeHint().height())
         self.setFixedHeight(self.shapeOptionsBox.sizeHint().height() + self.fieldsBox.sizeHint().height() + self.unitsbox.sizeHint().height() + height)
 
     def show_cylinder_fields(self):
--- a/nexus_constructor/add_component_window.py
+++ b/nexus_constructor/add_component_window.py
@@ -65,6 +65,7 @@ class AddComponentDialog(Ui_AddComponentDialog):
         initial_edit: bool,
         nx_classes=None,
         tree_view_updater: Callable = None,
+        use_dialogs: bool = False
     ):
         self._tree_view_updater = tree_view_updater
         self._scene_widget = scene_widget
@@ -95,6 +96,7 @@ class AddComponentDialog(Ui_AddComponentDialog):
         self.valid_file_given = False
         self.pixel_options: PixelOptions = None
         self.setupUi()
+        self.use_dialogs = use_dialogs
 #        self.setModal(True)
         self.setWindowModality(Qt.WindowModal)
         if self.initial_edit:
@@ -114,6 +116,9 @@ class AddComponentDialog(Ui_AddComponentDialog):
             self._group_parent.children.remove(self._group_container.group)
 
     def _confirm_cancel(self) -> bool:
+        if not self.use_dialogs:
+            return True
+
         quit_msg = "Do you want to close the group editor?"
         reply = QMessageBox.question(
             self,
--- a/nexus_constructor/add_component_window.py
+++ b/nexus_constructor/add_component_window.py
@@ -97,7 +97,6 @@ class AddComponentDialog(Ui_AddComponentDialog):
         self.pixel_options: PixelOptions = None
         self.setupUi()
         self.use_dialogs = use_dialogs
-#        self.setModal(True)
         self.setWindowModality(Qt.WindowModal)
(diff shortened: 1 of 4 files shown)


ui/add_component.py
```diff
@@ -1,26 +1,28 @@
...
-class Ui_AddComponentDialog(QDialog):
+class Ui_AddComponentDialog(QWidget):
     def __init__(self, parent: QtWidgets.QWidget, container: GroupContainer):
         super().__init__(parent)
         self._group_container = container
 
     def setupUi(self):
         self.setObjectName("AddComponentDialog")
-        self.resize(1600, 900)
+#        self.resize(1600, 900)
```
mattclarke, line 16:
```suggestion
```

nexus_constructor/add_component_window.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         initial_edit: bool,
         nx_classes=None,
         tree_view_updater: Callable = None,
+        use_dialogs: bool = False,
```
mattclarke, line 68: The plan is not use dialogs any more?
If so, just remove the code rather than sticking it behind a flag.

ggoneiESS: There have been a couple of changes that I've made that will be useful for a user to self-select:
 - the camera setting
 - confirmation dialogs

A user may want to have the confirmation when cancelling, particularly for long edits to a component.  The code for the options widget seemed a bit out of scope of this ticket.  So is it still best to delete even if we are keeping it for future implementation?

mattclarke: YAGNI.

nexus_constructor/add_component_window.py
```diff
@@ -128,11 +132,13 @@ def _confirm_cancel(self) -> bool:
 
     def _cancel_new_group(self):
         if self._confirm_cancel():
+            self._rejected()
```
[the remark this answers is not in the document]
ggoneiESS: I have reintroduced the bug where multiple panes appear. I don't think the relevant code is in this part but in mainwindow.py .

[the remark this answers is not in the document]
mattclarke: I didn't pick this piece of code specifically. I just needed somewhere to report it and was too lazy to do a proper comment :D

nexus_constructor/add_component_window.py
ggoneiESS: removed :)

nexus_constructor/add_component_window.py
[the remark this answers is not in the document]
ggoneiESS: had to check the attribute existed again but it's a cleaner fix than initially (no try catch)

nexus_constructor/main_window.py
```diff
@@ -240,6 +240,10 @@ def _update_3d_view_with_component_shapes(self):
             self.sceneWidget.add_transformation(component)
 
     def show_add_component_window(self, group: Group, new_group: bool):
+        if hasattr(self, "add_component_window") and not self.add_component_window.isHidden():
+            self.add_component_window._rejected()
+            self.add_component_window.setHidden(True)
+
```
mattclarke, line 246: This is just a variation of the hack you were unhappy with last week.
It is also just papering over the cracks.

The question to ponder is: why are we creating a new AppComponentDialog every time?

ggoneiESS: I see the point - the answer is because each component is different in terms of parent node, type etc.  You can see in :120 when we edit a group the difference is we pass a different flag which states whether we use empty groups or existing groups.  It doesn't simply 'fill in' the dialog/window etc, but creates a new object that has many members initialised with particular values which are important for various methods in the dialog (for example, remembering where the component goes after it is made, or remembering the initial state of the group so it can be restored if the user cancels it).  However, there are various object inside the AddComponentDialog class, such as the main UI layout, that mean it makes sense to have as a separate part (rather than as part of this MainWindow class).  Finally, the dialog is deleted afterwards so we are not getting a memory leak or anything like that.

I've removed the hasattr check, and now we have have a clean and consistent way of handling the dialog

mattclarke: This is not clean.

Repeated clicks of the edit button each create a new AppComponentDialog which are possibly not destructed (I see the memory usage increase on my machine at least).

It should not be necessary to construct a new AppComponentDialog every time just to pass some values into it, i.e. add a method something like `def set_initial_values(group, is_new_group)`.

nexus_constructor/main_window.py
```diff
@@ -60,6 +60,7 @@ def setupUi(self, main_window):
         self.component_tree_view_tab.set_up_model(self.model)
         self._update_views()
         self.simple_tree_view.triggered.emit()
+        self.add_component_window = QWidget()
```
mattclarke, line 63: This is where you should create the AppComponentDialog.

